PR #473, opened 2014-11-27
Adds footer to signed contract.

haarts:
boom

### DIFF
--- a/src/core/Contract.cpp
+++ b/src/core/Contract.cpp
@@ -1381,7 +1381,8 @@ bool Contract::ParseRawFile()
                 continue;
             }
 
-            if (bHaveEnteredContentMode && line.find("END SIGNED") != std::string::npos) {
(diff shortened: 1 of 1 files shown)


lclc:
@haarts @OttoAllmendinger:  How did you proceed in the Go implementation?

Do we want to vote on this? This PR is getting dusty ;)

OttoAllmendinger:
We have `ReadIncomplet()` and `WriteIncomplete()` on the go side already, so this is unnecessary.

If @haarts wants to make this work so that the function returns `false` if there is no signature section, we could spend a few more on this, but I'd rather just close it.

benma:
I'd keep it open, the Go side is significantly messier than it needs to be.

benma:
> <ttll> don't forget that skipping over lines also has consequences 
> <ttll> the displayed contract != what is hashed

With that, I see that this thing is too much effort. Closing.

PR #506, opened 2014-12-19
ZeroMQ: enable Curve

toxeus:


OttoAllmendinger:
nice :+1:

OttoAllmendinger:
Good to see that the tests pass. 

Do you think we could do it minimally secure? Read the keys from the command line/config/environment?

toxeus:
@OttoAllmendinger the correct way would be to fetch the keys from the notary provider contract. I don't think it's really worth the effort to bridge the current lack of the aforementioned mechanism with cmdline/config/environment because OTEnvelop "security" is still here. Keep in mind that we would have also to adapt the opentxs-tests and the Android client.

OttoAllmendinger:
Oops, didn't notice that OTEnvelope is still in there. Nevermind then, we can figure out how to do it properly with the NotaryContract and leave it like that. I'll merge on Monday.

toxeus:
@OttoAllmendinger please don't merge yet. The CI machine still needs some packages to be installed.

toxeus:
@OttoAllmendinger you're free to merge

OttoAllmendinger:
feels good man

PR #510, opened 2014-12-23
Fix mem leak in loading nym

mhamrle:
GetOrLoadNym returns pointer to Nym but caller should not delete the object. The owner is OTWallet. When loading public nyms the Nym object was not stored and caller did not delete it so the public nym leaked. Now OTWallet has two lists, one with private nyms (that is also stored) and one with with public nym.


mhamrle:
rebased

benma:
:+1:, merging.

PR #520, opened 2015-01-08
client/server: forward declare zsock_t 

toxeus:
This way we can encapsulate czmq.h inside the cpp file which minimizes
the conflict introduced by czmq.h including inttypes.h which prohibits
cinttypes from being properly included (PRIxYZ macros are not defined).


toxeus:
@mhamrle please review

PR #523, opened 2015-01-09
Remove the "<?xml version=..." prologs.

benma:
It is optional per XML spec, provides no meaningful information for OT,
and makes parsing harder.
Also, the version was set inconsistently (often with invalid XML version numbers).

### DIFF
--- a/include/opentxs/client/OTAPI.hpp
+++ b/include/opentxs/client/OTAPI.hpp
@@ -580,7 +580,6 @@ public:
     ASSET CONTRACTS will want to make sure they put something like this at the
     top:
 
-    <?xml version="1.0"?>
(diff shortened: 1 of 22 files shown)


OttoAllmendinger:
For context, see [MEP-3](https://github.com/monetas/protocol-docs/blob/master/content/mep/mep-003.md)

> * XML declarations are written only once at the top. This means that embedded XML has to avoid writing it.

I don't remember why we didn't choose to drop them completely. Maybe we should revise the MEP?

benma:
I don't know if there ever was a discussion about dropping them, only that they should not appear nested in the the armored parts.

In my opinion, we can just drop them altogether.

PR #534, opened 2015-01-15
CMake: fix shared lib extension for OSX

toxeus:


deps/CMakeLists.txt
```diff
@@ -88,7 +88,11 @@ add_subdirectory(czmq)
 # Create czmq_local for use inside opentxs' CMake project. Since czmq_local is
 # marked as IMPORTED, CMake will not complain about it not being exported.
 add_library(czmq_local SHARED IMPORTED GLOBAL)
-set_property(TARGET czmq_local PROPERTY IMPORTED_LOCATION ${CMAKE_BINARY_DIR}/lib/libczmq.so)
+set(lib_ext so)
+if (APPLE)
+  set(lib_ext dylib)
+endif()
+set_property(TARGET czmq_local PROPERTY IMPORTED_LOCATION ${CMAKE_BINARY_DIR}/lib/libczmq.${lib_ext})
```
mhamrle: This is not working on windows. What about use this:
```
get_target_property(CZMQ_FULLNAME czmq "LOCATION")
set_property(TARGET czmq_local PROPERTY IMPORTED_LOCATION ${CZMQ_FULLNAME})
```

toxeus: @mhamrle I know that this doesn't work on Windows. But I didn't want to create a fix that I cannot test. Anyway, I did not know about the `LOCATION` property. But my CMake 3.1.0 discourages this approach.
From `cmake --help-policy CMP0026`:

> CMP0026
> -------

> Disallow use of the LOCATION target property.

> CMake 2.8.12 and lower allowed reading the LOCATION target
property (and configuration-specific variants) to
determine the eventual location of build targets.  This relies on the
assumption that all necessary information is available at
configure-time to determine the final location and filename of the
target.  However, this property is not fully determined until later at
generate-time.  At generate time, the $\<TARGET_FILE\> generator
expression can be used to determine the eventual LOCATION of a target
output.

> Code which reads the LOCATION target property can be ported to use the
$\<TARGET_FILE\> generator expression together with the file(GENERATE)
subcommand to generate a file containing the target location.

> The OLD behavior for this policy is to allow reading the LOCATION
properties from build-targets.  The NEW behavior for this policy is to
not to allow reading the LOCATION properties from build-targets.

> This policy was introduced in CMake version 3.0.  CMake version
3.1.0 warns when the policy is not set and uses OLD behavior.  Use
the cmake_policy command to set it to OLD or NEW explicitly.

What do you think?

mhamrle: @toxeus I have cmake 2.8, I did not know about CMP0026.
I'll merge it.

PR #541, opened 2015-01-23
Export currency decimal power and currency factor

mhamrle:


toxeus:
@mhamrle can you please add to the commit message what this is good for?

mhamrle:
@toxeus  Done. Android client needs to limit the amount of decimal digits a user can enter.

toxeus:
@mhamrle thx! And why must they limit the amount of decimal digits?

mhamrle:
@toxeus User experience. For example in contract BTC the smallest amount you can use is 0.001 BTC. If user does not know about this limitation he can set smaller amount and UI have to check it and inform user that he is not able to set so small amount (or amount with so many decimal digits). When you can check it during entering value you save lots of user stories and user displeasures.

toxeus:
@mhamrle makes absolute sense. Thank you!

PR #542, opened 2015-01-23
fix typo

mhamrle:


lclc:
:+1: 
Maybe we should scan the source code with a text writing application / LibreOffice ;)

PR #543, opened 2015-01-25
cmake: fix linking problem with cmake version <3.0

lclc:
Easy backwards compatibility cmake vs <3

http://www.cmake.org/cmake/help/v3.0/command/target_link_libraries.html

With using LINK_PUBLIC and LINK_PRIVATE instead of just PUBLIC/PRIVATE (like we already have it in the core lib) it will still work with cmake versions below 3.
And above 3 aswell.

In general I agree to have it as requirement CMake 3. But since we don't use any features so far from there I think this fix is a good idea.

(I just recognized it because my system replaced my manually installed CMake 3.x with a CMake 2.8.x version with an update)

### DIFF
--- a/src/client/CMakeLists.txt
+++ b/src/client/CMakeLists.txt
@@ -151,7 +151,8 @@ else()
   )
 endif()
 
-target_link_libraries(opentxs-client PUBLIC opentxs-ext opentxs-cash opentxs-basket opentxs-core PRIVATE czmq_local)
+target_link_libraries(opentxs-client LINK_PUBLIC opentxs-ext opentxs-cash opentxs-basket opentxs-core)
+target_link_libraries(opentxs-client LINK_PRIVATE czmq_local)
 set_lib_property(opentxs-client)
 
 if(WIN32)
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -42,9 +42,11 @@ else()
 endif()
 
 if(WIN32)
-  target_link_libraries(opentxs-server PUBLIC opentxs-ext opentxs-cash opentxs-basket opentxs-core irrxml PRIVATE czmq_local)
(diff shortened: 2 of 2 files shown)


toxeus:
@lclc thank you for the valuable contributions! 

I don't understand what this commit fixes. Our CI runs on latest Ubuntu LTS that ships with [cmake 2.8.12](http://packages.ubuntu.com/trusty/cmake) and builds opentxs fine. Also the Docker image is based on the same Ubuntu version and also builds fine. 

You can easily verify that with Docker (make sure to use version 1.4.x):
```bash
docker build -t opentxs $PATH_TO_OPENTXS_REPO
```

So, I'd really like to learn what the exact problem is we're having. Can you please explain (ideally in the commit message) what this commit is fixing? And also don't forget to fix the commit message's title ;)

lclc:
I have cmake version 2.8.11.2, I just assumed it was a 3.0 change, but it seems to be introduced with 2.8.12 (sorry^^): (introduces PRIVATE and PUBLIC):
http://www.cmake.org/pipermail/cmake/2013-October/056020.html

Since those don't exist in 2.8.11 I get a linker error (he's looking for -lPRIVATE and -lPUBLIC).
So we can either change those back to LINK_PUBLIC & LINK_PRIVATE or make it consistent and use only the new PRIVATE and PUBLIC everywhere (in core we still use the old way: https://github.com/Open-Transactions/opentxs/blob/develop/src/core/CMakeLists.txt#L121).

Do we want to define 2.8.12 as the minimum CMake version?
I guess we should list our dependencies (independent from the OS) and the minimal version somewhere in the docs folder (or in the README.md)

toxeus:
@lclc in that case I'd change `cmake_minimum_required(VERSION 2.8.11)` to `cmake_minimum_required(VERSION 2.8.12)`.

lclc:
Ok, I'll do that, update my cmake to 3.1.1 and change the style in the CMake core file.

PR #546, opened 2015-01-26
CMake: set minimum version to 2.8.12, change linking style

lclc:
Discussed here: https://github.com/Open-Transactions/opentxs/pull/543#issuecomment-71439726


toxeus:
@lclc Great! Thx!

PR #550, opened 2015-02-02
Export methods for accessing currency properties.

mhamrle:
Android NDK does not support locales. So it is better if amount
formating is be done in Java.

Closing monetas/opentxs#40


toxeus:
@mhamrle please rebase

FellowTraveler:
I'm merging this. Martin is correct that standard locale stuff is not working. Therefore I will add options for the comma / decimal symbols to be passed into the function. Simplificator needs this for demo version.

PR #557, opened 2015-02-05
no more need to erase PID by hand on UNIX (client only)

FellowTraveler:
I can add this fix on the server side as well. But for now client was higher priority.

### DIFF
--- a/src/client/OpenTransactions.cpp
+++ b/src/client/OpenTransactions.cpp
@@ -198,6 +198,36 @@ extern "C" {
 }
 #endif
 
+// On certain platforms we can actually check to
+// see if the PID is running, and if not, we can
+// proceed even if the PID is in the file. (To
+// spare the user from having to delete the pid
+// file by hand.)
+//
+#ifdef PREDEF_PLATFORM_UNIX
+
+#ifdef __APPLE__
+#include "TargetConditionals.h"
+#include <CoreServices/CoreServices.h>
+#endif
+
+#if defined(ANDROID)
+// blank
+#elif defined(TARGET_OS_MAC)
+
+#if TARGET_OS_MAC
+#include <sys/wait.h>
+#define OT_CHECK_PID 1
+#endif
+
+#else
+#include <sys/types.h>
+#include <sys/wait.h>
+#define OT_CHECK_PID 1
+#endif
+
+#endif
+
 #define CLIENT_CONFIG_KEY "client"
 #define CLIENT_MASTER_KEY_TIMEOUT_DEFAULT 300
 #define CLIENT_WALLET_FILENAME "wallet.xml"
@@ -378,6 +408,7 @@ public:
     void OpenPid(const String& strPidFilePath);
     void ClosePid();
     bool IsPidOpen() const;
+    static bool PIDAutorecoverImpossible(uint32_t pid);
 
 private:
     bool m_bIsPidOpen;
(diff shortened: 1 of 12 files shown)


toxeus:

toxeus:

FellowTraveler:

toxeus:

FellowTraveler:

toxeus:

src/client/OpenTransactions.cpp
```diff
@@ -444,8 +488,13 @@ void OT_API::Pid::OpenPid(const String& strPidFilePath)
...
             pid_infile.close();
 
-            // There was a real PID in there.
-            if (old_pid != 0) {
+// There was a real PID in there.
+#ifdef OT_CHECK_PID
+            if ((old_pid != 0) && DoIKnowIfPidIsRunning(old_pid))
+#else
+            if (old_pid != 0)
+#endif
```
toxeus: This `#ifdef`s is not necessary. You should call `DoIKnowIfPidIsRunning(old_pid)` in every case and handle the `#ifdef` inside `DoIKnowIfPidIsRunning(old_pid)` (as you already do). But it should return true by default.

src/client/OpenTransactions.cpp
```diff
@@ -395,6 +426,19 @@ OT_API::Pid::~Pid()
...
+{
+#ifdef OT_CHECK_PID
+    while (waitpid(-1, 0, WNOHANG) > 0) {
+        // Wait for defunct....
+    }
+
+    if (0 == kill(pid, 0)) return true; // Process exists
+#endif
+    return false;
+}
```
toxeus: Better:

```c++
bool OT_API::Pid::DoIKnowIfPidIsRunning(uint32_t pid)
{
#ifdef OT_CHECK_PID
    while (waitpid(-1, 0, WNOHANG) > 0) {
          // Wait for defunct....
    }

    if (0 != kill(pid, 0)) return false; // Process exists
#endif
     return true;
}
```

See other comment.

src/client/OpenTransactions.cpp
FellowTraveler: It cannot return true by default, because by default, it does NOT know if the PID is actually a running process. Basically it would be saying "I know it is running" even when it isn't running, and it does not know, and can not know.

src/client/OpenTransactions.cpp
FellowTraveler: See other reply. Why return true if it really doesn't know if the PID is running, and in fact the PID may very well NOT be running? Seems like true is a lie.

FellowTraveler: Just to be clear: if it returns true, then you could have a situation where the process is NOT running twice, yet it still refuses to let you run since it THINKS it's running (when it's not.)

src/client/OpenTransactions.cpp
toxeus: If `OT_CHECK_PID` is not defined then `if (old_pid != 0)` is executed. It is equivalent to `if ((old_pid != 0) && true)` which means we could leave `DoIKnowIfPidIsRunning(old_pid)` if it would always return true for the `OT_CHECK_PID` not defined case. And this is exactly what my refactoring suggestion for `DoIKnowIfPidIsRunning(old_pid)` in the comment above does.

FellowTraveler: Okay lemme walk through this. Let's say it's Windows, so OT_CHECK_PID is not defined. Let's say the program is not running twice, and furthermore I do not know if it is or not. Then I call DoIKnowIfPIDRunning and it returns true.

So now it's NOT running, and the function cannot KNOW whether it's running, yet the function claims it IS running and that it KNOWS it is running.

I think you are right about the logic, however. At least, in the specific place where it's being called. But what if the function is ever used in any other place? It seems deceptive. I would think it should be renamed, at least, so it's not lying about its return value.

toxeus: Well, then let's rename it :ghost:

FellowTraveler: Since you are the one suggesting this alteration, I will give you the honor of naming the function.

toxeus: I propose `PIDAutorecoverImpossible` as the new function name.

FellowTraveler:

toxeus:

toxeus:
@FellowTraveler the machine running the CI crashed. @grealish is trying to fix it

weissjeffm:
retest this please

PR #566, opened 2015-03-02
Generate transportKey and write it to the notaryProviderContract. Remove transport encryption based on OTEnvelope.

benma:
This PR puts a `<transportKey>` element into the notaryProviderContract, containing the notary's public transport key. Using [zcert_public_txt](http://czmq.zeromq.org/manual:zcert) returns the [Z85 encoding](http://rfc.zeromq.org/spec:32) of the key. Despite the Z85 RFC explicitly mentioning XML in the reasoning of the encoding, the encoding contains the `<>` characters. This is why I also base64-encode it. @toxeus, @OttoAllmendinger, do you have a better solutions in mind? I did not yet investigate PEM or section markers, I figured @toxeus is the expert on that.

The public and the private key are stored in `server_data/credentials/<serverNymID>/transportKey` (and `_secret`). The notary generates/reads this. The client reads the transport public key from the notary provider contract.

### DIFF
--- a/include/opentxs/core/OTServerContract.hpp
+++ b/include/opentxs/core/OTServerContract.hpp
@@ -134,6 +134,7 @@
 #define OPENTXS_CORE_OTSERVERCONTRACT_HPP
 
 #include "Contract.hpp"
+#include <czmq.h>
 
 namespace opentxs
 {
@@ -149,6 +150,8 @@ public:
     EXPORT virtual ~OTServerContract();
 
     EXPORT bool GetConnectInfo(String& strHostname, int32_t& nPort) const;
+    EXPORT unsigned char* GetTransportKey() const;
+    static zcert_t* LoadOrCreateTransportKey(const String& nymID);
     EXPORT virtual void CreateContents(); // Only used when first generating an
                                           // asset or server contract. Meant for
                                           // contracts which never change after
@@ -166,6 +169,7 @@ protected:
     String m_strHostname;
     int32_t m_nPort;
     String m_strURL;
+    unsigned char* m_transportKey;
 };
 
 } // namespace opentxs
--- a/src/client/OTServerConnection.cpp
+++ b/src/client/OTServerConnection.cpp
@@ -154,7 +154,8 @@ namespace opentxs
 // but either way the connections need a pointer to the wallet
 // they are associated with, so they can access those accounts.
 OTServerConnection::OTServerConnection(OTClient* theClient,
-                                       const std::string& endpoint)
+                                       const std::string& endpoint,
+                                       const unsigned char* transportKey)
     : socket_zmq(zsock_new_req(NULL))
     , m_pNym(nullptr)
     , m_pServerContract(nullptr)
@@ -165,10 +166,10 @@ OTServerConnection::OTServerConnection(OTClient* theClient,
         OT_FAIL;
     }
     zsock_set_linger(socket_zmq, 1000);
+    // Set new client public and secret key.
     zcert_apply(zcert_new(), socket_zmq);
-    // test key value taken from `man zmq_curve`
-    zsock_set_curve_serverkey(socket_zmq,
-                              "rq:rM>}U?@Lns47E1%kR.o@n%FcmmsL/@{H8]yf7");
+    // Set server public key.
+    zsock_set_curve_serverkey_bin(socket_zmq, transportKey);
 
     if (zsock_connect(socket_zmq, "%s", endpoint.c_str())) {
         Log::vError("Failed to connect to %s\n", endpoint.c_str());
--- a/src/client/OTServerConnection.cpp
+++ b/src/client/OTServerConnection.cpp
@@ -222,10 +222,8 @@ void OTServerConnection::send(OTServerContract* pServerContract, Nym* pNym,
     const Nym* pServerNym = pServerContract->GetContractPublicNym();
     OT_ASSERT(nullptr != pServerNym);
 
-    OTEnvelope theEnvelope;
-    String strEnvelopeContents;
-    theMessage.SaveContractRaw(strEnvelopeContents);
-    theEnvelope.Seal(*pServerNym, strEnvelopeContents);
+    String strContents;
+    theMessage.SaveContractRaw(strContents);
 
     otOut << "\n=====>BEGIN Sending " << theMessage.m_strCommand
           << " message via ZMQ... Request number: "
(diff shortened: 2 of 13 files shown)


OttoAllmendinger:
About Z85: I think it's supposed to be "easier" to embed because you only need to escape the `&<>` characters, but you have to do some escaping nonetheless. 

Maybe we should just base64-encode the raw binary key data and skip the Z85 encoding altogether?

benma:
I did that initially, then I saw that `zsock_set_curve_serverkey(sock, key)` (as used before) took the Z85-encoded key, and I did not find an encoding function in czmq after a quick search. So I made my life easy ;) MIght might give this another try, though, because the binary is also shorter.

benma:
Oh, and Twin Peaks is the best show of all time.

src/core/OTServerContract.cpp
```diff
@@ -265,10 +297,23 @@ int32_t OTServerContract::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
...
+    }
+    if (!strcmp("transportKey", xml->getNodeName())) {
+        if (!SkipToTextField(xml)) {
+            return -1;
+        }
+        const char* transportKeyB64 = xml->getNodeData();
+        size_t outLen;
+        m_transportKey = reinterpret_cast<char*>(
+            OTCrypto::It()->Base64Decode(transportKeyB64, &outLen, false));
+        if (outLen != 41) {
```
mhamrle: Can we use constant instead of magic number 41?

OttoAllmendinger:
It's probably this one with the `_bin` suffix:

```
CZMQ_EXPORT void zsock_set_curve_serverkey_bin (void *self, const byte *curve_serverkey);
```

> Oh, and Twin Peaks is the best show of all time.

src/client/OTServerConnection.cpp
```diff
@@ -240,7 +239,7 @@ void OTServerConnection::send(OTServerContract* pServerContract, Nym* pNym,
            << "\n\n";
 }
 
-bool OTServerConnection::send(const OTEnvelope& theEnvelope)
+bool OTServerConnection::send(const String& theEnvelope)
```
mhamrle: `theEnvelope` is not envelope now, the variable should be renamed.

src/core/OTServerContract.cpp
benma: Fixed.

benma:
> It's probably this one with the `_bin` suffix:

You were right. Fixed.

src/client/OTServerConnection.cpp
benma: Fixed.

benma:
test this please

mhamrle:
:+1:

weissjeffm:
test this please

deps/czmq
```diff
@@ -1 +1 @@
-Subproject commit 869c3145b2ae60802f6239b9615b2bcd8b715d96
+Subproject commit 9c24aea3defe84cd72d7e6ac1845b6215f3d7aad
```
toxeus: Please don't downgrade to the `v3.0.0` tag. I have checked out 869c3145b2ae60802f6239b9615b2bcd8b715d96 for a reason: the `CZMQ_BUILD_TESTS` variable has been introduced which we are [using to disable the czmq tests](https://github.com/Open-Transactions/opentxs/blob/develop/deps/CMakeLists.txt#L62-L63).

src/ext/CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
 # Copyright (c) Monetas AG, 2014
 
+include_directories(SYSTEM
+  ${ZEROMQ_INCLUDE_DIRS}
+  ${CZMQ_INCLUDE_DIR}
+)
+
+
```
toxeus: Why is this necessary when you did not touch any files in `src/ext`?

include/opentxs/core/OTServerContract.hpp
```diff
@@ -166,6 +169,7 @@ class OTServerContract : public Contract
     String m_strHostname;
     int32_t m_nPort;
     String m_strURL;
+    unsigned char* m_transportKey;
```
toxeus: I don't know if this type is forced by czmq but I'd prefer `uint8_t`.

toxeus:

deps/czmq
benma: That was not deliberate. I don't know what's wrong with this submodule, they always show up as changed in `git status` (along with another submodule). Will fix tomorrow.

include/opentxs/core/OTServerContract.hpp
benma: "forced" is the wrong word in a language where you can do reinterpret_cast ;) Anyway, the czmq functions do work with `unsigned char*`, and I don't mind it because this is the convention for binary data (as opposed to `char*` for strings). If you insist on `uint8_t`, I can change it, but I see no reason why it is better. Why do you prefer it?

src/ext/CMakeLists.txt
benma: If I remember correctly, OTPayment depends on one of the core/* classes where I include czmq.h. I will double-check tomorrow.

benma: OTPayment.hpp -> InstantiateContract.hpp -> OTServerContract -> czmq.h.

src/core/CMakeLists.txt
```diff
@@ -118,8 +123,8 @@ else()
   )
 endif()
 
-target_link_libraries(opentxs-core PRIVATE opentxs-recurring opentxs-script opentxs-cron opentxs-trade otprotob irrxml bitcoin-base58)
-target_link_libraries(opentxs-core PUBLIC ${PROTOBUF_LIBRARY} ${OPENSSL_LIBRARIES} ${ZLIB_LIBRARY} ${OPENTXS_SYSTEM_LIBRARIES} ${keyring})
+target_link_libraries(opentxs-core PRIVATE opentxs-recurring opentxs-script opentxs-cron opentxs-trade otprotob irrxml bitcoin-base58 czmq_local)
```
toxeus: You don't need to put `czmq_local` also here.

include/opentxs/core/OTServerContract.hpp
toxeus: Let's not use `reinterpret_cast`. Just wanted to check if it's czmq that uses that convention. [More info](http://stackoverflow.com/questions/16138237/when-is-uint8-t-%E2%89%A0-unsigned-char/16138470#16138470)

src/core/CMakeLists.txt
benma: Where? It is there, in both `target_link_libraries` statements.

toxeus:

toxeus:

toxeus:
:+1: excellent

PR #568, opened 2015-03-04
Normalize the data that is signed/verified by trimming the spaces.

benma:
as part of the paylaod.

### DIFF
--- a/include/opentxs/core/Contract.hpp
+++ b/include/opentxs/core/Contract.hpp
@@ -160,6 +160,8 @@ class OTSignature;
 typedef std::list<OTSignature*> listOfSignatures;
 typedef std::map<std::string, Nym*> mapOfNyms;
 
+String trim(const String& str);
+
 class Contract
 {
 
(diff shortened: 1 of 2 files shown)


toxeus:

weissjeffm:
This broke market tests. http://ci.opentransactions.org:8080/job/opentxs-PR-functional-test/9/

I'll look into why jenkins didn't mark it failed here on github.

toxeus:

OttoAllmendinger:
One thing that we have to look out for is that we only do this where spaces really are unimportant, so probably all XML strings or armored strings that OT sign. But even though the the function is called `SignContract` and takes `string`, I can imagine where OT uses this to sign binary data (grep shows ~500 call sites). `Trim` would be a bad idea in those cases. 

So we should double-check this. or do `trim` only where we are absolutely sure that no binary stuff gets trimmed.

benma:
@OttoAllmendinger 

You are probably right, and we should do the trimming in Contract.SignContract/VerifyContract. I'll look into it.

benma:
retest this please

benma:
retest this please

benma:
retest this please

toxeus:
:+1: great job!

PR #579, opened 2015-03-12
Dockerfile: add Dockerfile for Android

toxeus:
This PR decouples the Android and the Linux Dockerfiles. This is only possible because Docker learned the `-f` flag in version 1.5. Then, we can use the Android Dockerfile to re-enable Android builds on Jenkins.


mhamrle:
:+1:

PR #581, opened 2015-03-13
Dockerfile.android: add CMD for easy lib fetch

toxeus:
Now, the Android libs can easily be built and fetched:

```bash
docker build --pull --rm --force-rm -t android -f Dockerfiles/Dockerfile.android .
LIBS_DIR=$HOME/android_libs
mkdir $LIBS_DIR
docker run --rm -v $LIBS_DIR:/mnt/ android
ls $LIBS_DIR
```

**IMPORTANT**: this *only* works with Docker verion >= 1.5.


mhamrle:
:+1:

PR #607, opened 2015-07-08
Remove unused const (build error fix)

faizkhan00:
This fixes a build issue on Linux (possibly other platforms) where a build error is triggered due to a unused const in Log.cpp - This comments out the line for build purposes, but also allows developers to uncomment it in case it is used later.


faizkhan00:
Closing because this doesn't appear to be an issue on Ubuntu, just ArchLinux

PR #628, opened 2015-11-09
Correct Identifier calculations

justusranvier:
Conversions from OTData to OTPassword are lossy except for very small OTData sizes.

Add an implementation of OpenSSL::Digest() that accepts OTData


FellowTraveler:
Just a warning to any users: This fix changes how IDs are calculated for contracts, and you will have to regenerate any test data you are using.

PR #644, opened 2015-12-21
Implement Contact credentials

justusranvier:
Contact credentials are non-key credentials that allow users to associate arbitrary public information with their nym.

### DIFF
--- a/include/opentxs/core/crypto/NymParameters.hpp
+++ b/include/opentxs/core/crypto/NymParameters.hpp
@@ -62,9 +62,9 @@ public:
 
     void setNymParameterType(NymParameterType theKeytype);
 
-    Credential::CredentialType credentialType() const;
+    proto::CredentialType credentialType() const;
 
-    void setCredentialType(Credential::CredentialType theCredentialtype);
+    void setCredentialType(proto::CredentialType theCredentialtype);
 
     inline proto::SourceType SourceType() const
     {
@@ -95,7 +95,7 @@ public:
 
     NymParameters(
         NymParameterType theKeytype,
-        Credential::CredentialType theCredentialtype);
+        proto::CredentialType theCredentialtype);
     NymParameters() = default;
     ~NymParameters() = default;
 
@@ -136,13 +136,13 @@ private:
 
 #if defined(OT_CRYPTO_SUPPORTED_KEY_SECP256K1)
     NymParameterType nymType_ = NymParameterType::SECP256K1;
-    Credential::CredentialType credentialType_ = Credential::HD;
+    proto::CredentialType credentialType_ = proto::CREDTYPE_HD;
 #elif defined(OT_CRYPTO_SUPPORTED_KEY_RSA)
     NymParameterType nymType_ = NymParameterType::LEGACY;
-    Credential::CredentialType credentialType_ = Credential::LEGACY;
+    proto::CredentialType credentialType_ = proto::CREDTYPE_LEGACY;
 #else
     NymParameterType nymType_ = NymParameterType::ERROR;
-    Credential::CredentialType credentialType_ = Credential::ERROR_TYPE;
+    proto::CredentialType credentialType_ = proto::CREDTYPE_ERROR;
 #endif
 
 //----------------------------------------
--- a/include/opentxs/core/crypto/NymParameters.hpp
+++ b/include/opentxs/core/crypto/NymParameters.hpp
@@ -39,6 +39,8 @@
 #ifndef OPENTXS_CORE_CRYPTO_NYMPARAMETERS_HPP
 #define OPENTXS_CORE_CRYPTO_NYMPARAMETERS_HPP
 
+#include <memory>
+
 #include <opentxs-proto/verify/VerifyCredentials.hpp>
 
 #include <opentxs/core/crypto/Credential.hpp>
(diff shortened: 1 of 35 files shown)


include/opentxs/core/crypto/NymParameters.hpp
```diff
@@ -86,6 +88,13 @@ class NymParameters
         sourceProofType_ = sType;
     }
 
+    inline proto::ContactData ContactData() const
+    {
+        return *contact_data_;
```
FellowTraveler: If proto::ContactData has a move constructor, we might want to return std::move( *contact_data_ );

src/client/OpenTransactions.cpp
```diff
@@ -4488,6 +4490,109 @@ Nym* OT_API::GetOrLoadNym(const Identifier& NYM_ID, bool bChecking,
...
+    return fromNym.ContactData();
+}
+
+OT_API::ClaimSet OT_API::GetClaims(const Nym& fromNym) const
+{
+    proto::ContactData data = GetContactData(fromNym);
+
+    OT_API::ClaimSet claimSet;
+
+    for (auto& section: data.section()) {
```
FellowTraveler: Probably could do const auto & for these.

src/core/crypto/ContactCredential.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+namespace opentxs
+{
+
+ContactCredential::ContactCredential(
+    CredentialSet& parent,
+    const proto::Credential& credential)
+        : ot_super(parent, credential)
+{
+    m_strContractType = "CONTACT CREDENTIAL";
```
FellowTraveler: Take this out.

src/core/crypto/CredentialSet.cpp
```diff
@@ -265,8 +225,11 @@ CredentialSet::CredentialSet(
             LoadChildKeyCredential(it);
         }
     } else {
+        Credential* masterCred = Credential::CredentialFactory(
+            *this,
+            serializedCredentialSet.mastercredential());
         m_MasterCredential.reset(
-            new MasterCredential(*this, serializedCredentialSet.mastercredential()));
+            dynamic_cast<MasterCredential*>(masterCred));
```
FellowTraveler: Animal * pAnimal = Animal::Factory(TYPE_CAT);
Cat * pCat = dynamic_cast<Cat *>(pAnimal);
if (nullptr != pAnimal && nullptr == pCat)
   delete pAnimal;
pAnimal = nullptr;
if (pCat)
{

}

src/core/crypto/CredentialSet.cpp
```diff
@@ -458,14 +420,31 @@ bool CredentialSet::ReEncryptPrivateCredentials(
...
+        otErr << __FUNCTION__
+              << ": Could not parse retrieved credential as a protobuf.\n";
+        return false;
+    }
+
+    Credential* purported =
+        Credential::CredentialFactory(
+            *this, *serializedCred, proto::CREDROLE_MASTERKEY);
+
+    m_MasterCredential.reset(dynamic_cast<MasterCredential*>(purported));
```
FellowTraveler: Same thing here.
If you KNOW for SURE that it's the right type, then just force it with a static_cast. It can't return null, so you will know for sure if the factory returned null.

src/core/crypto/CredentialSet.cpp
```diff
@@ -532,13 +511,11 @@ bool CredentialSet::Load_Master(const String& strNymID,
...
-    if (!validProto) {
-        otErr << __FUNCTION__ << ": Invalid serialized master credential.\n";
-        return false;
-    }
+    Credential* purported =
+        Credential::CredentialFactory(
+            *this, *serializedCred, proto::CREDROLE_MASTERKEY);
 
-    m_MasterCredential.reset(new MasterCredential(*this, *serializedCred));
+    m_MasterCredential.reset(dynamic_cast<MasterCredential*>(purported));
```
FellowTraveler: Same here.
Another way I do this, is I use a unique_ptr of the base type, and then IF and ONLY IF the dynamic_cast succeeds, then I release the unique_ptr into a subclass unique_ptr.

src/core/crypto/CredentialSet.cpp
```diff
@@ -566,8 +542,11 @@ bool CredentialSet::LoadChildKeyCredentialFromString(const String& strInput,
         m_mapCredentials.erase(it);
     }
 
-    ChildKeyCredential* pSub = new ChildKeyCredential(*this, strInput);
-    std::unique_ptr<ChildKeyCredential> theSubAngel(pSub);
+    serializedCredential credential =
+        Credential::ExtractArmoredCredential(strInput);
+    Credential* pSub =
+        Credential::CredentialFactory(*this, *credential);
+    std::unique_ptr<Credential> theSubAngel(pSub);
```
FellowTraveler: This illustrates what I just described above. Mostly.

src/core/crypto/KeyCredential.cpp
```diff
@@ -77,7 +77,7 @@ bool KeyCredential::VerifySignedBySelf() const
 {
     OT_ASSERT(m_SigningKey);
 
-    serializedSignature publicSig = GetSelfSignature(false);
+    serializedSignature publicSig = SelfSignature(false);
```
FellowTraveler: Do that boolean typedef trick.

src/core/crypto/MasterCredential.cpp
```diff
@@ -222,27 +186,63 @@ MasterCredential::~MasterCredential()
...
 }
 
+bool MasterCredential::Verify(const Credential& credential) const
+{
+    serializedCredential serializedCred =
+        credential.asSerialized(
+            Credential::AS_PUBLIC,
+            Credential::WITHOUT_SIGNATURES);
+
+    if (!proto::Verify(*serializedCred, credential.Role(), false)) {
```
FellowTraveler: what's the false?

PR #654, opened 2016-01-08
New storage system

justusranvier:
Add opentxs::App as a singleton class for handing functionality common to both client and server apps.

Currently App handles CryptoEngine and the new opentxs::Storage system.

opentxs::Storage saves and retrieves serialized protobuf object from a key-value store.

Two key-value stores are currently implemented: filesystem and sqlite3.


include/opentxs/core/app/App.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+//Singlton class for providing an interface to process-level resources.
+class App
+{
+private:
+    static App* instance_pointer_;
+
+    Storage* storage_ = nullptr;
+    CryptoEngine* crypto_ = nullptr;
+    std::thread* periodic_thread_;
```
FellowTraveler: needs:

 =nullptr;

include/opentxs/core/app/Settings.hpp
```diff
@@ -39,19 +39,19 @@
...
 private:
-    OTSettings(const OTSettings&);
-    OTSettings& operator=(const OTSettings&);
+    Settings(const Settings&);
+    Settings& operator=(const Settings&);
 
-    class OTSettingsPvt;
-    OTSettingsPvt* pvt;
+    class SettingsPvt;
+    SettingsPvt* pvt;
```
FellowTraveler: =nullptr;

include/opentxs/core/crypto/Credential.hpp
```diff
@@ -162,12 +162,13 @@ class Credential : public Contract
     // Inherited from opentxs::Contract
     EXPORT virtual void CalculateContractID(Identifier& newID) const;
     virtual void ReleaseSignatures(const bool onlyPrivate);
-    virtual bool SaveContract();
-    virtual bool SaveContract(const char* szFoldername, const char* szFilename);
+    using ot_super::SaveContract;
```
FellowTraveler: How does this know which version of SaveContract from ot_super that it's using?

Does it just infer based on the fact that the version with no parameters is provided (override below) ?

include/opentxs/core/crypto/CryptoEngine.hpp
```diff
@@ -119,7 +119,7 @@ class CryptoEngine
     EXPORT Bip32& BIP32();
 #endif
 
-    EXPORT static CryptoEngine& Instance();
+    EXPORT static CryptoEngine& Factory();
```
FellowTraveler: A factory returns a new instance every single time, like this:

EXPORT static CryptoEngine * Factory();

Based on our conversation, rename this to "It" or "Instance" or whatever.

include/opentxs/core/util/OTFolders.hpp
```diff
@@ -128,7 +129,6 @@ class OTFolders
     EXPORT static const String& Nymbox();
     EXPORT static const String& Outbox();
     EXPORT static const String& PaymentInbox();
-    EXPORT static const String& Pubcred();
     EXPORT static const String& Pubkey();
```
FellowTraveler: Looks like Pubkey is unused. Well, it's used in one function, but I don't see THAT function being called anywhere either.

Grep output:

~/Projects/opentxs $ fgrep -nR "OTFolders::Pubkey" ./src/*
./src/core/Nym.cpp:2569:    return OTDB::Exists(OTFolders::Pubkey().Get(), pstrID->Get());
./src/core/util/OTFolders.cpp:253:const String& OTFolders::Pubkey()
~/Projects/opentxs $ ox ./src/core/Nym.cpp
~/Projects/opentxs $ 
~/Projects/opentxs $ 
~/Projects/opentxs $ 
~/Projects/opentxs $ fgrep -nR Server_PubKeyExists ./src/*
./src/core/Nym.cpp:2555:bool Nym::Server_PubKeyExists(String* pstrID) // Only used
~/Projects/opentxs $ 
~/Projects/opentxs $ 
~/Projects/opentxs $ ..
~/Projects $ cd opentxs-notary
~/Projects/opentxs-notary $ fgrep -nR Server_PubKeyExists ./src/*
~/Projects/opentxs-notary $ ..
~/Projects $

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+
+typedef std::function<bool(const uint32_t, const std::string&, std::string&)>
+    Digest;
+typedef std::function<std::string()>
+    Random;
+
+template<class T>
+std::string ProtoAsString(const T& serialized)
+{
+    int size = serialized.ByteSize();
```
FellowTraveler: should be auto size

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+    char* protoArray = new char [size];
```
FellowTraveler: This could have been:

char* protoArray = new char [size]{};

Which would initialize to zero.

But you actually probably don't want to do that here, it would only slow things down.

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+// The interface to a particular KV database is provided by child classes
+// implementing this interface. Implementations need only provide methods for
+// storing/retrieving arbitrary key-value pairs, and methods for setting and
+// retrieving the hash of the root index object.
+//
+// The implementation of this interface must support the concept of "buckets"
+// Objects are either stored and retrieved from either the primary bucket, or
+// the alternate bucket. This allows for garbage collection of outdated keys
+// to be implemented.
+class Storage
```
FellowTraveler: Might want to put these function implementations in a CPP file, and put that in the same folder as the HPP file, and include the CPP file at the bottom using #include ""

FellowTraveler: But FYI it is basically fine how it is now, no need to change at this point.

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+{
+template<class T>
+bool LoadProto(
+    const std::string hash,
```
FellowTraveler: Any reason this isn't a reference?

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+    std::shared_ptr<T>& serialized)
+{
+    if (hash.empty()) { return false; }
```
FellowTraveler: Just put a log here saying, "Hash was empty, are you running for the first time?"

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+
+    std::unique_lock<std::mutex> llock(location_lock_);
+    bool attemptFirst = gc_running_ ? !alt_location_ : alt_location_;
```
FellowTraveler: You said might want to rename alt_location to current_bucket.  Whatever name works, just the current name makes the logic look backwards.

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+
+            foundInSecondary = Verify(*serialized);
+        }
+    }
+
+    return (foundInPrimary || foundInSecondary);
+}
+
+template<class T>
+bool StoreProto(const T data)
```
FellowTraveler: Might want a reference here.

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+{
+    if (nullptr != digest_) {
+        std::string plaintext = opentxs::ProtoAsString<T>(data);
```
FellowTraveler: plaintext should probably be const.

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    return (foundInPrimary || foundInSecondary);
+}
+
+template<class T>
+bool StoreProto(const T data)
+{
+    if (nullptr != digest_) {
```
FellowTraveler: Should probably log if digest_ IS null.  (It works if not null, but it doesn't log if IS null.)

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+        std::string plaintext = opentxs::ProtoAsString<T>(data);
+        std::string key;
+        digest_(Storage::HASH_TYPE, plaintext, key);
+
+        return Store(
+            key,
+            plaintext,
+            alt_location_);
```
FellowTraveler: current bucket

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+    }
+    return false;
+}
+
+private:
+    static Storage* instance_pointer_;
+    static const uint32_t GC_INTERVAL;
```
FellowTraveler: Should add this GC_INTERVAL to the config file.

Also, suggest changing uint32_t to int64_t since the Settings object uses that type already.

Here's an example (latency) from the config file:
```
[latency]
linger = 1000
send_timeout = 1000
recv_timeout = 10000
```

Here's the code in OpenTransactions.cpp:
```
    {
        int64_t lValue; bool bIsNewKey;
        p_Config->CheckSet_long("latency", "recv_timeout",
                                OTServerConnection::getRecvTimeout(), lValue, bIsNewKey);
        OTServerConnection::setRecvTimeout(static_cast<int>(lValue));
    }
```

Therefore you could add this to the client and server cfg files:
```
[garbage_collect]
interval = 3600
```

And add this code:
```
    {
        int64_t lValue; bool bIsNewKey; // These are both for output.
        p_Config->CheckSet_long("garbage_collect", "interval",
                                opentxs::Storage::defaultGcInterval(), lValue, bIsNewKey);
        opentxs::Storage::setGcInterval(lValue);
    }
```

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+    bool UpdateRoot();
+
+    void CollectGarbage();
+    bool MigrateKey(const std::string& key);
+
+    Storage(Storage const&) = delete;
+    Storage& operator=(Storage const&) = delete;
+
+protected:
+    const uint32_t HASH_TYPE = 2; // BTC160
```
FellowTraveler: We don't have a #define we can use here instead of a hardcoded 2 ?

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+    Digest digest_ = nullptr;
+    Random random_ = nullptr;
```
FellowTraveler: No need to set these to nullptr since they are both a std::function, which is similar to std::shared_ptr -- it has its own constructor and it already sets its internal pointer to nullptr on construction.

Basically you never have to set an OBJECT to null, only a raw pointer. And these aren't raw pointers.

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+
+    std::mutex init_lock_; // controls access to Read() method
+    std::mutex cred_lock_; // ensures atomic writes to credentials_
+    std::mutex nym_lock_; // ensures atomic writes to nyms_
+    std::mutex write_lock_; // ensure atomic writes
+    std::mutex gc_lock_; // prevents multiple garbage collection threads
+    std::mutex location_lock_; // ensures atomic updates of alt_location_
+    std::mutex bucket_lock_; // ensures buckets not changed during read
+
+    std::string root_ = "";
```
FellowTraveler: Probably don't need to initialize the strings here, suggest removing the quotes and see if it still works. Surprised if it doesn't.

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+    virtual bool Store(
+        const std::string& key,
+        const std::string& value,
+        const bool altLocation) = 0;
+    virtual bool EmptyBucket(const bool altLocation) = 0;
+
+public:
+    // Factory method for instantiating the singleton. param is a child
+    // class-defined instantiation parameter.
+    static Storage& Factory(
```
FellowTraveler: A Factory should return a pointer to a new instance every single time it's called.

If it's a singleton, recommend renaming it to It or Instance or something.

include/opentxs/storage/Storage.hpp
```diff
@@ -0,0 +1,237 @@
...
+        const Digest& hash,
+        const Random& random,
+        const std::string& param = "");
+
+    bool Load(
+        const std::string id,
```
FellowTraveler: Should probably be a reference here, but depends on whether or not you're constructing it from other types in this spot. I doubt that though.

include/opentxs/storage/StorageFS.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+// Simple filesystem implementation of opentxs::storage
+class StorageFS : public Storage
+{
+private:
+    typedef Storage ot_super;
+
+    friend Storage;
+
+    std::string folder_ = "";
```
FellowTraveler: already initialized in constructor. No need to initialized unless with a real value.

include/opentxs/storage/StorageSqlite3.hpp
```diff
@@ -0,0 +1,120 @@
...
+// SQLite3 implementation of opentxs::storage
+class StorageSqlite3 : public Storage
+{
+private:
+    typedef Storage ot_super;
+
+    friend Storage;
+
+    std::string folder_ = "";
+    ::sqlite3* db_;
```
FellowTraveler: Probably no need to initialize folder string.

Probably no need to use scope resolution operator for sqlite3.

db_ should be initialized here to nullptr. It's just a pointer to a struct and needs to be null here.

include/opentxs/storage/StorageSqlite3.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+    static const std::string primaryTable;
+    static const std::string secondaryTable;
+    static const std::string controlTable;
+    static const std::string rootKey;
+
+    static std::string GetTableName(const bool altLocation)
+    {
+        return altLocation ?
```
FellowTraveler: current bucket?

include/opentxs/storage/StorageSqlite3.hpp
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    StorageSqlite3() = delete;
+    // param is interpreted to mean a full path to the folder where the
+    // database file should be stored
+    StorageSqlite3(
+        const std::string& param,
+        const Digest& hash,
+        const Random& random);
+    StorageSqlite3(StorageSqlite3 const&) = delete;
```
FellowTraveler: This const is correct, but probably not the one you wanted to use here.

Probably wanted const StorageSqlite3 &

include/opentxs/storage/StorageSqlite3.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+    bool Select(
+        const std::string& key,
+        const std::string& tablename,
+        std::string& value);
+    bool Upsert(
+        const std::string& key,
+        const std::string& tablename,
+        const std::string& value);
+    bool Create(const std::string& tablename);
```
FellowTraveler: Should be named createTable?

src/core/Nym.cpp
```diff
@@ -3069,6 +3015,14 @@ bool Nym::LoadCredentialIndex(const String& armoredIndex)
     return true;
 }
 
+bool Nym::LoadCredentialIndex(const String& armoredIndex)
```
FellowTraveler: This armoredIndex parameter should definitely be an OTASCIIArmor&, NOT an OTString&.

Unless it's bookended, in which case, you take it as a String& parameter, and then do ascBlah.LoadFromBookendedString(). I don't remember the exact func name but it's something like that.

Then in the places where you are USING LoadCredentialIndex(), it looks like this:

In OTClient.cpp:
```
    const String publicNym(theReply.m_ascPayload.Get());
    Nym theTargetNym;
    theTargetNym.LoadCredentialIndex(publicNym);
```

Becomes:
```
    Nym theTargetNym;
    theTargetNym.LoadCredentialIndex(theReply.m_ascPayload);
```

Nym.hpp:220:    EXPORT const String asPublicNym() const;

SHOULD BE:    EXPORT const OTASCIIArmor asPublicNym() const;

in Contract.cpp:
```
                const String publicNym = theSigner.asPublicNym();

                std::unique_ptr<Nym> pNym(new Nym);
                pNym->LoadCredentialIndex(publicNym);
```

Becomes:
```
                const OTASCIIArmor publicNym = theSigner.asPublicNym();

                std::unique_ptr<Nym> pNym(new Nym);
                pNym->LoadCredentialIndex(publicNym);
```

Etc.

src/core/Nym.cpp
```diff
@@ -4553,11 +4507,9 @@ bool Nym::VerifyTransactionStatementNumbersOnNym(Nym& THE_NYM) // THE_NYM is
 // static
 bool Nym::DoesCertfileExist(const String& strNymID)
```
FellowTraveler: This function can be entirely removed.

src/core/OTServerContract.cpp
```diff
@@ -177,13 +181,17 @@ void OTServerContract::CreateContents()
 
     // Write the transportKey
     const Nym* nym = m_mapNyms["signer"];
+
+    if (nullptr == nym) {
+        otErr << __FUNCTION__ << "Contract has no signer nym." << std::endl;
```
FellowTraveler: Might want to return if this fails.

Which means might have to return bool instead of void.

Not important right now, I'm assuming this is a low level function that is only called when there's definitely going to be a signer anyway. But if that's actually the case, then we should ASSERT here.

src/core/app/App.cpp
```diff
@@ -0,0 +1,152 @@
...
+            const std::string&,
+            std::string&)>(&CryptoHash::Digest),
+        &(Crypto().Hash()),
+        std::placeholders::_1,
+        std::placeholders::_2,
+        std::placeholders::_3);
+
+    Random random = std::bind(&CryptoUtil::RandomFilename, &(Crypto().Util()));
+
+    std::shared_ptr<OTDB::StorageFS> storage;
```
FellowTraveler: These could both be on the same line:

```
 std::shared_ptr<OTDB::StorageFS> storage(OTDB::StorageFS::Instantiate());
```

src/core/app/App.cpp
```diff
@@ -0,0 +1,152 @@
...
+    storage.reset(OTDB::StorageFS::Instantiate());
+    std::string root_path = OTFolders::Common().Get();
+    std::string path;
+
+#ifdef OT_STORAGE_FS
```
FellowTraveler: Can move this to storage subclass.

src/core/app/App.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+void App::Periodic()
+{
+    for (;;) {
+        // Collect garbage, if necessary
+        if (nullptr != storage_) {
+            storage_->RunGC();
+        }
+
+        Log::SleepMilliseconds(250);
```
FellowTraveler: Might want to make the 250 configurable in the settings.

Anytime we have a hardcoded number we should ask ourselves if it should be in the config file (likely should.)

src/core/app/App.cpp
```diff
@@ -0,0 +1,152 @@
...
+{
+    if (nullptr == instance_pointer_)
+    {
+        instance_pointer_ = new App;
+    }
+
+    return *instance_pointer_;
+}
+
+CryptoEngine& App::Crypto() const
```
FellowTraveler: Weird that it lets you return a non-const member from a const method.

Makes me feel like CryptoEngine MUST be typedef'd as const, for this to even compile.

Same with the next function.

src/core/app/App.cpp
```diff
@@ -0,0 +1,152 @@
...
+Storage& App::DB() const
+{
+    OT_ASSERT(nullptr != storage_)
+
+    return *storage_;
+}
+
+void App::Cleanup()
+{
+    storage_->Cleanup();
```
FellowTraveler: Probably can just put the Cleanup call in the destructors here.

src/core/app/App.cpp
```diff
@@ -0,0 +1,152 @@
...
+    crypto_->Cleanup();
+
+    delete storage_;
+    delete crypto_;
```
FellowTraveler: Should set these to nullptr here.

src/core/app/App.cpp
```diff
@@ -0,0 +1,152 @@
...
+}
+
+App::~App()
+{
+    if ((nullptr != periodic_thread_) && periodic_thread_->joinable()) {
+        periodic_thread_->join();
+        delete periodic_thread_;
```
FellowTraveler: Need to set this to nullptr here.

It's probably not used again after this point, but why worry?

src/core/crypto/CryptoHash.cpp
```diff
@@ -55,6 +57,27 @@ bool CryptoHash::Digest(
     return Digest(hashType, plaintext, digest);
 }
 
+bool CryptoHash::Digest(
+    const uint32_t type,
+    const std::string& data,
+    std::string& encodedDigest)
+{
+    OTData plaintext(data.c_str(), data.size() + 1); // +1 for null terminator
```
FellowTraveler: Remove the +1 here.

You might do this if you were copying INTO a string, since you want to make sure you add the null terminator, but even then, if you were copying into std::string it would add the null terminator itself, inside the constructor.

src/core/crypto/CryptoHash.cpp
```diff
@@ -55,6 +57,27 @@ bool CryptoHash::Digest(
...
+        static_cast<CryptoHash::HashType>(type),
+        plaintext,
+        result);
+
+    if (success) {
+        encodedDigest.assign(
+            App::Me().Crypto().Util().Base58CheckEncode(result).Get());
+    }
+
+    return false;
```
FellowTraveler: Should be return success?

src/core/util/OTFolders.cpp
```diff
@@ -246,10 +250,6 @@ const String& OTFolders::PaymentInbox()
 {
     return GetFolder(s_strPaymentInbox);
 }
-const String& OTFolders::Pubcred()
-{
-    return GetFolder(s_strPubcred);
-}
 const String& OTFolders::Pubkey()
```
FellowTraveler: Delete Pubkey, no longer used.

src/core/util/OTFolders.cpp
```diff
@@ -166,8 +168,6 @@ bool OTFolders::GetSetAll()
     if (!GetSetFolderName(config, KEY_PAYMENTINBOX, DEFAULT_PAYMENTINBOX,
                           s_strPaymentInbox))
         return false;
-    if (!GetSetFolderName(config, KEY_PUBCRED, DEFAULT_PUBCRED, s_strPubcred))
-        return false;
     if (!GetSetFolderName(config, KEY_PUBKEY, DEFAULT_PUBKEY, s_strPubkey))
```
FellowTraveler: No longer used.

src/server/OTServer.cpp
```diff
@@ -383,15 +302,102 @@ void OTServer::Init(bool readOnly)
...
+            std::string defaultContract =
+            "<notaryProviderContract version=\"2.0\">\n\n";
+            defaultContract += "<entity shortname=\"localhost\"\n";
+            defaultContract += " longname=\"Localhost Test Contract\"\n";
+            defaultContract += " email=\"serverfarm@blahcloudcomputing.com\"\n";
+            defaultContract += " serverURL=\"https://blahtransactions.com/vers/1/\"/>\n\n";
+            defaultContract += "<notaryServer hostname=\"localhost\"\n";
+            defaultContract += " port=\"7085\"\n";
+            defaultContract += " URL=\"https://blahtransactions.com/vers/1/\" />\n\n";
+            defaultContract += "</notaryProviderContract>\n";
```
FellowTraveler: What you did here, will work fine.

But FYI you could have done this:

```
defaultContract +=
"<entity shortname=\"localhost\"\n"
" longname=\"Localhost Test Contract\"\n"
" email=\"serverfarm@blahcloudcomputing.com\"\n"
" serverURL=\"https://blahtransactions.com/vers/1/\"/>\n\n"
"<notaryServer hostname=\"localhost\"\n"
" port=\"7085\"\n"
" URL=\"https://blahtransactions.com/vers/1/\" />\n\n"
"</notaryProviderContract>\n"
```

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+#ifdef OT_STORAGE_FS
+#include <opentxs/storage/StorageFS.hpp>
+#elif defined OT_STORAGE_SQLITE
+#include <opentxs/storage/StorageSqlite3.hpp>
+#endif
+
+namespace opentxs
+{
+Storage* Storage::instance_pointer_ = nullptr;
+const uint32_t Storage::GC_INTERVAL = 60 * 60 * 1; // hourly
```
FellowTraveler: Suggest replacing hardcoded values with this:  OT_TIME_HOUR_IN_SECONDS

(Defined in Common.hpp)

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+Storage::Storage(const Digest& hash, const Random& random)
+{
+    isLoaded_ = false;
```
FellowTraveler: I would think the isLoaded_ = false should be in the header, not here.

If it WAS here, I'd do it like this:

```
Storage::Storage(const Digest& hash, const Random& random) : isLoaded_(false)
{
    isLoaded_ = false;
```

(But like I said, it would be even better in the header, and not here at all.)

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+    Init(hash, random);
+}
+
+void Storage::Init(const Digest& hash, const Random& random)
+{
+    digest_ = hash;
+    random_ = random;
+}
+
+Storage& Storage::Factory(
```
FellowTraveler: I think this function should probably look like this:

```
Storage& Storage::It(
    const Digest& hash,
    const Random& random,
    const std::string& param)
{
    if (nullptr == instance_pointer_)
    {
#ifdef OT_STORAGE_FS
    instance_pointer_ = new StorageFS(param, hash, random);
#elif defined OT_STORAGE_SQLITE
    instance_pointer_ = new StorageSqlite3(param, hash, random);
#else
#error  "blah blah blah"
#endif
    }
    assert(nullptr != instance_pointer_);

    return *instance_pointer_;
}
```

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        root_ = LoadRoot();
+
+        if (root_.empty()) { return; }
+
+        std::shared_ptr<proto::StorageRoot> root;
+
+        if (!LoadProto<proto::StorageRoot>(root_, root)) { return; }
+
+        items_ = root->items();
+        alt_location_ = root->altlocation();
```
FellowTraveler: current bucket

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+                    it.hash()));
+            }
+        }
+    }
+}
+
+bool Storage::UpdateNymCreds(const std::string& id, const std::string& hash)
+{
+    // Reuse existing object, since it may contain more than just creds
+    if ((!id.empty()) && (!hash.empty())) {
```
FellowTraveler: Can remove some parens here

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        }
+    }
+
+    return false;
+}
+
+bool Storage::UpdateCredentials(const std::string& id, const std::string& hash)
+{
+    // Do not test for existing object - we always regenerate from scratch
+    if ((!id.empty()) && (!hash.empty())) {
```
FellowTraveler: you can remove the extra parens.

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+        // Block reads while updating credential map
```
FellowTraveler: // Block writes while searching credential map.

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        std::unique_lock<std::mutex> credlock(cred_lock_);
+        credentials_.insert(std::pair<std::string, std::string>(id, hash));
+        proto::StorageCredentials credIndex;
+        credIndex.set_version(1);
+        for (auto& cred : credentials_) {
+            if (!cred.first.empty() && !cred.first.empty()) {
```
FellowTraveler: This seems like an obvious bug.

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+bool Storage::UpdateNyms(const proto::StorageNym& nym)
+{
+    // Do not test for existing object - we always regenerate from scratch
+    if (nullptr != digest_) {
+        std::string id = nym.nymid();
+        std::string plaintext = ProtoAsString<proto::StorageNym>(nym);
+        std::string hash;
+        digest_(Storage::HASH_TYPE, plaintext, hash);
+
+        // Block reads while updating nym map
```
FellowTraveler: Block writes.

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        std::unique_lock<std::mutex> nymLock(nym_lock_);
+        nyms_.insert(std::pair<std::string, std::string>(id, hash));
+        proto::StorageNymList nymIndex;
+        nymIndex.set_version(1);
+        for (auto& nym : nyms_) {
+            if (!nym.first.empty() && !nym.first.empty()) {
```
FellowTraveler: Another obvious bug, copy pasted from above.

FellowTraveler: BTW if this "should never happen" then probably want to assert here.

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        }
+    }
+
+    return false;
+}
+
+bool Storage::UpdateNyms(const proto::StorageNym& nym)
+{
+    // Do not test for existing object - we always regenerate from scratch
+    if (nullptr != digest_) {
```
FellowTraveler: if (digest_)

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+    std::shared_ptr<proto::StorageItems> items;
+
+    if (!LoadProto<proto::StorageItems>(items_, items)) {
+        items = std::make_shared<proto::StorageItems>();
+        items->set_version(1);
+    } else {
+        items->clear_creds();
+    }
+
+    if (nullptr != digest_) {
```
FellowTraveler: if (digest_)

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+    std::shared_ptr<proto::StorageItems> items;
+
+    if (!LoadProto<proto::StorageItems>(items_, items)) {
+        items = std::make_shared<proto::StorageItems>();
+        items->set_version(1);
+    } else {
+        items->clear_nyms();
+    }
+
+    if (nullptr != digest_) {
```
FellowTraveler: if (digest_)

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        root = std::make_shared<proto::StorageRoot>();
+        root->set_version(1);
+        root->set_altlocation(false);
+        std::time_t time = std::time(nullptr);
+        root->set_lastgc(static_cast<int64_t>(time));
+    } else {
+        root->clear_items();
+    }
+
+    if (nullptr != digest_) {
```
FellowTraveler: if (digest_)

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        std::string plaintext = ProtoAsString<proto::StorageItems>(items);
+        std::string hash;
+        digest_(Storage::HASH_TYPE, plaintext, hash);
+
+        items_ = hash;
+
+        root->set_version(1);
+        root->set_items(hash);
+
+        assert(Verify(*root));
```
FellowTraveler: I wonder if we shouldn't move OT_ASSERT and OT_ASSERT_MSG into another library. So we can use it here. So on Android, we can get this stack trace when this asserts.

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+        if (StoreProto<proto::StorageRoot>(*root)) {
+            plaintext = ProtoAsString<proto::StorageRoot>(*root);
+            digest_(Storage::HASH_TYPE, plaintext, hash);
+
+            root_ = hash;
```
FellowTraveler: root_hash_

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+    }
+
+    return false;
+}
+
+bool Storage::UpdateRoot(
+    proto::StorageRoot& root,
+    const std::string& gcroot)
+{
+    if (nullptr != digest_) {
```
FellowTraveler: if (digest_)

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        root.set_altlocation(alt_location_);
```
FellowTraveler: current bucket

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+        std::time_t time = std::time(nullptr);
+        last_gc_ = static_cast<int64_t>(time);
+        root.set_lastgc(last_gc_);
+
+        root.set_gcroot(gcroot);
+
+        assert(Verify(root));
```
FellowTraveler: assert again, might want to add above comment here as well with a todo.

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+    proto::StorageRoot& root,
+    const std::string& gcroot)
+{
+    if (nullptr != digest_) {
+        root.set_altlocation(alt_location_);
+
+        std::time_t time = std::time(nullptr);
+        last_gc_ = static_cast<int64_t>(time);
+        root.set_lastgc(last_gc_);
+
```
FellowTraveler: You're supposed to be setting a root.bool here.

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        root.set_gcroot(gcroot);
+
+        assert(Verify(root));
+
+        if (StoreProto<proto::StorageRoot>(root)) {
+            std::string hash;
+            std::string plaintext = ProtoAsString<proto::StorageRoot>(root);
+            digest_(Storage::HASH_TYPE, plaintext, hash);
+
+            root_ = hash;
```
FellowTraveler: root_hash_ ?

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+// this version is for ending garbage collection only
+bool Storage::UpdateRoot()
+{
+    std::shared_ptr<proto::StorageRoot> root;
+
+    bool loaded = LoadProto<proto::StorageRoot>(root_, root);
+
+    assert(loaded);
+
+    if (nullptr != digest_) {
```
FellowTraveler: if (digest_)

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        gc_running_ = false;
+        root->set_gc(false);
+
+        assert(Verify(*root));
```
FellowTraveler: might wanna put todo comment here regarding OT_ASSERT

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+        if (StoreProto<proto::StorageRoot>(*root)) {
+            std::string hash;
+            std::string plaintext = ProtoAsString<proto::StorageRoot>(*root);
+            digest_(Storage::HASH_TYPE, plaintext, hash);
+
+            root_ = hash;
```
FellowTraveler: root_hash_ ?

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+            return StoreRoot(hash);
+        }
+    }
+
+    return false;
+}
+
+bool Storage::Load(
+    const std::string id,
```
FellowTraveler: should be reference?

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+    std::shared_ptr<proto::Credential>& cred)
+{
+    if (!isLoaded_) { Read(); }
+
+    bool found = false;
+    std::string hash = "";
```
FellowTraveler: Probably don't need the = "";

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+    }
+
+    return false;
+}
+
+bool Storage::Load(
+    const std::string id,
+    std::shared_ptr<proto::Credential>& cred)
+{
+    if (!isLoaded_) { Read(); }
```
FellowTraveler: This is safe, but only because Read() is checking isLoaded_ AGAIN, after it locks.

Otherwise I'd suggest that isLoaded_ should probably be std::atomic

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+    bool found = false;
+    std::string hash = "";
+
+    // block writes while searching credential map
+    std::unique_lock<std::mutex> credlock(cred_lock_);
+    auto it = credentials_.find(id);
+    if (it != credentials_.end()) {
```
FellowTraveler: This is fine, but could have been:

```
found = (it != credentials_.end());

if (found)
    hash = it->second;
```

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+    }
+    credlock.unlock();
+
+    if (found) { return LoadProto<proto::Credential>(hash, cred); }
+
+    return false;
+}
+
+bool Storage::Load(
+    const std::string id,
```
FellowTraveler: Need a reference here probably

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+    std::shared_ptr<proto::CredentialIndex>& credList)
+{
+    if (!isLoaded_) { Read(); }
```
FellowTraveler: This is only safe because Read() checks isLoaded_ AGAIN after locking. Otherwise isLoaded_ should be atomic here.

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+    bool found = false;
+    std::string nymHash = "";
```
FellowTraveler: No need to initialize this.

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+    // block writes while searching nym map
+    std::unique_lock<std::mutex> nymLock(nym_lock_);
+    auto it = nyms_.find(id);
```
FellowTraveler: Could have been:

```
found = (it != nyms_.end());

if (found)
        nymHash = it->second;
```

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+            }
+        }
+    }
+
+    return false;
+}
+
+bool Storage::Store(const proto::Credential& data)
+{
+    if (!isLoaded_) { Read(); }
```
FellowTraveler: Only safe because Read() checks again after locking.
Otherwise need isLoaded_ to be atomic

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+    }
+
+    if (existingPrivate && (proto::KEYMODE_PRIVATE != data.mode())) {
+        std::cout << "Skipping update of existing private credential with "
+                  << "non-private version." << std::endl;
+
+        return true;
+    }
+
+    if (nullptr != digest_) {
```
FellowTraveler: if (digest_)

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        if (savedCredential) {
+            return UpdateCredentials(data.id(), key);
+        }
+    }
+    return false;
+}
+
+bool Storage::Store(const proto::CredentialIndex& data)
+{
+    if (!isLoaded_) { Read(); }
```
FellowTraveler: isLoaded_ would have to be atomic, but it's safe since Read checks it AGAIN after locking

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+    if (nullptr != digest_) {
```
FellowTraveler: if (digest_)

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        std::string key;
+        digest_(Storage::HASH_TYPE, plaintext, key);
+
+        std::lock_guard<std::mutex> writeLock(write_lock_);
+        bool saved = Store(
+            key,
+            plaintext,
+            alt_location_);
+
+        if (saved) {
```
FellowTraveler: brackets are unnecessary here, no big deal though.

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        }
+    }
+
+    return false;
+}
+
+void Storage::CollectGarbage()
+{
+    std::unique_lock<std::mutex> llock(location_lock_);
+    bool oldLocation = alt_location_;
```
FellowTraveler: alt location is current bucket

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+    alt_location_ = !alt_location_;
+    llock.unlock();
+
+    std::shared_ptr<proto::StorageRoot> root;
+    // Do not allow changes to root index object until we've updated it.
+    std::unique_lock<std::mutex> writeLock(write_lock_);
+    std::string gcroot = root_;
+    if (!LoadProto<proto::StorageRoot>(root_, root)) {
+        // If there is no root object, then there's nothing to gc
+        writeLock.unlock();
```
FellowTraveler: I don't think you need to release the lock here, since you are returning, the destructor likely already does that, might want to double check.

FellowTraveler: Alright, I looked, YES the destructor unlocks already.

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+        gc_running_ = false;
+        return;
+    }
+    std::string gcitems = root->items();
+    bool updated = UpdateRoot(*root, gcroot);
```
FellowTraveler: should be const bool

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+    return true; // the key must have already been in the active bucket
+}
+
+void Storage::RunGC()
+{
+    if (!isLoaded_) { return; }
+
+    std::lock_guard<std::mutex> gclock(gc_lock_);
+    std::time_t time = std::time(nullptr);
+    bool intervalExceeded = ((time - last_gc_) > Storage::GC_INTERVAL);
```
FellowTraveler: should be const bool

src/storage/Storage.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+    if (!gc_running_ && ( gc_resume_ || intervalExceeded)) {
+        assert (!gc_running_);
```
FellowTraveler: OT_ASSERT todo comment should go here

src/storage/StorageFS.cpp
```diff
@@ -0,0 +1,196 @@
...
+void StorageFS::Purge(const std::string& path)
+{
+    if (path.empty()) { return; }
+
+    boost::filesystem::remove_all(path);
+}
+
+std::string StorageFS::LoadRoot()
+{
+    if (folder_ != "") {
```
FellowTraveler: I think this works fine, and we do use it this way in some places.

But normally I'd do this:

```
if (!folder_.empty()) {
```

src/storage/StorageFS.cpp
```diff
@@ -0,0 +1,196 @@
...
+        std::string filename = folder_ + "/root";
```
FellowTraveler: Normally we have special sauce that inserts "/" or "\" depending on the OS.

Apparently forward slashes work now?  I guarantee that must be new, because they used to not work. (On Windows.)

src/storage/StorageFS.cpp
```diff
@@ -0,0 +1,196 @@
...
+            std::ios::in | std::ios::ate | std::ios::binary);
+
+        if (file.good()) {
+            std::ifstream::pos_type size = file.tellg();
+            file.seekg(0, std::ios::beg);
+
+            std::vector<char> bytes(size);
+            file.read(&bytes[0], size);
+
+            value.assign(&bytes[0], size);
```
FellowTraveler: This is a perfect example of where std::string adds the null terminator for you. That's why you're not doing a size+1 shit here.

src/storage/StorageFS.cpp
```diff
@@ -0,0 +1,196 @@
...
+            return true;
+        }
+    }
+
+    return false;
+}
+
+bool StorageFS::StoreRoot(const std::string& hash)
+{
+    if (folder_ != "") {
```
FellowTraveler: if (!folder_empty()) {

src/storage/StorageFS.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+            return true;
+        }
+    }
+
+    return false;
+}
+
+bool StorageFS::EmptyBucket(
+    const bool altLocation)
```
FellowTraveler: current bucket

src/storage/StorageFS.cpp
```diff
@@ -0,0 +1,196 @@
...
+{
+    assert(nullptr != random_);
```
FellowTraveler: SHOULD BE:   assert(random_);

(Since it's a function pointer object, not a raw pointer.)

Also should add a todo comment here re: OT_ASSERT

src/storage/StorageFS.cpp
```diff
@@ -0,0 +1,196 @@
...
+void StorageFS::Cleanup()
+{
+    ot_super::Cleanup();
+}
+
+StorageFS::~StorageFS()
+{
+    Cleanup();
+}
+
```
FellowTraveler:
```
class Storage
{
    void Cleanup_Storage()
    {
        // Do all the base class cleanup here.
    }

    virtual void Cleanup()
    {
        Cleanup_Storage();
    }

    virtual ~Storage()
    {
        Cleanup_Storage();
    }
};

class StorageFS
{
    void Cleanup_StorageFS()
    {
        // Do all the sub-class cleanup here.
    }

    void Cleanup() override
    {
        Cleanup_StorageFS();

        ot_super::Cleanup();
    }

    virtual ~StorageFS()
    {
        Cleanup_StorageFS();
    }
};

```

src/storage/StorageSqlite3.cpp
```diff
@@ -0,0 +1,196 @@
...
+    const std::string StorageSqlite3::controlTable = "control";
+    const std::string StorageSqlite3::rootKey = "root";
+
+StorageSqlite3::StorageSqlite3(
+    onst std::string& param,
+    const Digest&hash,
+    const Random& random)
+        : ot_super(hash, random)
+{
+    Init(param);
```
FellowTraveler: If Init is virtual then you definitely can't call it here, and have to do something similar to Cleanup() as shown above.

BUT the base class does NOT declare this version of Init, therefore you aren't actually calling a virtual function here, which is why it's probably fine.

src/storage/StorageSqlite3.cpp
```diff
@@ -0,0 +1,196 @@
...
+}
+
+bool StorageSqlite3::Select(
+    const std::string& key,
+    const std::string& tablename,
+    std::string& value)
+{
+    sqlite3_stmt *statement;
```
FellowTraveler: need =nullptr;

src/storage/StorageSqlite3.cpp
```diff
@@ -0,0 +1,196 @@
...
+    const std::string query =
+        "select v from " + tablename + " where k=?1;";
```
FellowTraveler: I think you need to put this character as quotes around the tablename:  `

FellowTraveler: Normally the data value needs quotes like this:  '

But since you are binding, you are safe here already.

FellowTraveler: Probably want to add the LIMIT statement as seen in this Moneychanger example:

```
    QString str_select = QString("SELECT * FROM `trade_archive` WHERE `receipt_id`=%1 LIMIT 0,1").arg(lReceiptID);
```

src/storage/StorageSqlite3.cpp
```diff
@@ -0,0 +1,196 @@
...
+}
+
+bool StorageSqlite3::Upsert(
+    const std::string& key,
+    const std::string& tablename,
+    const std::string& value)
+{
+    sqlite3_stmt *statement;
+    const std::string query =
+        "insert or replace into " + tablename +
```
FellowTraveler: need the back quote around table name:   `

src/storage/StorageSqlite3.cpp
```diff
@@ -0,0 +1,196 @@
...
+    sqlite3_finalize(statement);
+
+    return (result == SQLITE_DONE);
+}
+
+bool StorageSqlite3::Create(const std::string& tablename)
+{
+    const std::string createTable = "create table if not exists ";
+    const std::string tableFormat = " (k text PRIMARY KEY, v BLOB);";
+    const std::string sql = createTable + tablename + tableFormat;
```
FellowTraveler: + "`" + tablename + "`"

src/storage/StorageSqlite3.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    return (SQLITE_OK ==
+        sqlite3_exec(db_, sql.c_str(), nullptr, nullptr, nullptr));
+}
+
+bool StorageSqlite3::Purge(const std::string& tablename)
+{
+    const std::string sql = "DROP TABLE " + tablename + ";";
```
FellowTraveler: add single backquote around table name

src/storage/StorageSqlite3.cpp
```diff
@@ -0,0 +1,196 @@
...
+        SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE | SQLITE_OPEN_FULLMUTEX,
+        nullptr)) {
+            Create(StorageSqlite3::primaryTable);
+            Create(StorageSqlite3::secondaryTable);
+            Create(StorageSqlite3::controlTable);
+            sqlite3_exec(
+                db_, "PRAGMA journal_mode=WAL;", nullptr, nullptr, nullptr);
+    } else {
+        std::cout << "Failed to initialize database." << std::endl;
+        assert(false);
```
FellowTraveler: todo comment here for OT_ASSERT

src/storage/StorageSqlite3.cpp
```diff
@@ -0,0 +1,196 @@
...
+{
+    return Upsert(StorageSqlite3::rootKey, StorageSqlite3::controlTable, hash);
+}
+
+bool StorageSqlite3::Store(
+    const std::string& key,
+    const std::string& value,
+    const bool altLocation)
+{
+    return Upsert(key, GetTableName(altLocation), value);
```
FellowTraveler: current bucket

src/storage/StorageSqlite3.cpp
```diff
@@ -0,0 +1,196 @@
...
+}
+
+bool StorageSqlite3::EmptyBucket(const bool altLocation)
```
FellowTraveler: current bucket

src/storage/StorageSqlite3.cpp
```diff
@@ -0,0 +1,196 @@
...
+{
+    return Purge(GetTableName(altLocation));
+}
+
+void StorageSqlite3::Cleanup()
+{
+    sqlite3_close(db_);
+}
+
+StorageSqlite3::~StorageSqlite3()
```
FellowTraveler: Cleanup is definitely fucked.  See above code for fix a little ways up.

src/storage/StorageFS.cpp
FellowTraveler: One more thing, BTW:

Notice that the Cleanup override calls its own version, AND the ot_super version.

Whereas notice that the Destructor calls ONLY its own version, WITHOUT calling the ot_super version.

This is deliberate, because destructors are a little funny.

Any normal virtual method, overrides the base class version.

But a virtual destructor does NOT override the base class version! BOTH destructors are automatically called!

So then, why make the destructor virtual at all?

The reason is, because if you have a base pointer to a subclass, and the destructor isn't virtual, the subclass destructor will never get called!

Example:
```
Sub * psub = new Sub;

Base *pbase = psub;

delete pbase;   // Here, the subclass destructor will FAIL to be called unless the Base destructor is virtual!
```

Note that the base destructor gets called EITHER WAY, which is why you don't have to call it in the subclass destructor. Even though the destructor is virtual, both destructors are called, and there's no "override" (though the subclass's destructor will execute first.)

src/storage/StorageFS.cpp
justusranvier: "Note  File I/O functions in the Windows API convert "/" to "\" as part of converting the name to an NT-style name, except when using the "\\?\" prefix as detailed in the following sections."

https://msdn.microsoft.com/en-us/library/aa365247%28VS.85%29.aspx

PR #655, opened 2016-01-12
OpenDHT integration

justusranvier:
Add optional [OpenDHT](https://github.com/savoirfairelinux/opendht) integration for publishing and retrieving public contracts.

Currently insertion of server contracts, instrument definition contracts, and public nyms (credentials) is supported as well as fetching of public nyms and server contracts.

Depends on https://github.com/Open-Transactions/opentxs/pull/654

### DIFF
--- a/include/opentxs/core/app/App.hpp
+++ b/include/opentxs/core/app/App.hpp
@@ -42,6 +42,7 @@
 #include <thread>
 
 #include <opentxs/storage/Storage.hpp>
+#include <opentxs/core/app/Dht.hpp>
 #include <opentxs/core/crypto/CryptoEngine.hpp>
 
 namespace opentxs
@@ -55,6 +56,7 @@ private:
 
     Storage* storage_ = nullptr;
     CryptoEngine* crypto_ = nullptr;
+    Dht* dht_ = nullptr;
     std::thread* periodic_thread_ = nullptr;
     bool server_mode_ = false;
 
@@ -71,6 +73,7 @@ public:
 
     CryptoEngine& Crypto();
     Storage& DB();
+    Dht& DHT();
 
     void Cleanup();
     ~App();
--- a/include/opentxs/core/app/App.hpp
+++ b/include/opentxs/core/app/App.hpp
@@ -56,8 +56,10 @@ private:
     Storage* storage_ = nullptr;
     CryptoEngine* crypto_ = nullptr;
     std::thread* periodic_thread_ = nullptr;
+    bool server_mode_ = false;
 
-    App();
+    App(const bool serverMode);
+    App() = delete;
     App(const App&) = delete;
     App& operator=(const App&) = delete;
 
@@ -65,7 +67,7 @@ private:
     void Periodic();
 
 public:
-    static App& Me();
+    static App& Me(const bool serverMode = false);
 
     CryptoEngine& Crypto();
     Storage& DB();
--- a/include/opentxs/core/app/App.hpp
+++ b/include/opentxs/core/app/App.hpp
@@ -39,6 +39,7 @@
 #ifndef OPENTXS_CORE_APP_APP_HPP
 #define OPENTXS_CORE_APP_APP_HPP
 
+#include <limits>
 #include <thread>
 
 #include <opentxs/storage/Storage.hpp>
@@ -61,7 +62,10 @@ private:
     Storage* storage_ = nullptr;
 
     std::thread* periodic_thread_ = nullptr;
+
     bool server_mode_ = false;
+    int64_t last_nym_publish_ = 0;
+    int64_t nym_publish_interval_ = std::numeric_limits<int64_t>::max();
 
     App(const bool serverMode);
     App() = delete;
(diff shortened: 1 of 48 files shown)


include/opentxs/core/app/App.hpp
```diff
@@ -53,22 +56,38 @@ class App
...
+    Settings* config_ = nullptr;
     CryptoEngine* crypto_ = nullptr;
+    Dht* dht_ = nullptr;
+    Storage* storage_ = nullptr;
+
     std::thread* periodic_thread_ = nullptr;
 
-    App();
+    bool server_mode_ = false;
+    int64_t last_nym_publish_ = 0;
```
FellowTraveler: I recommend time64_t

src/core/app/App.cpp
```diff
@@ -99,20 +191,38 @@ void App::Periodic()
...
-App& App::Me()
+App& App::Me(const bool serverMode)
 {
     if (nullptr == instance_pointer_)
     {
-        instance_pointer_ = new App;
+        instance_pointer_ = new App(serverMode);
     }
 
     return *instance_pointer_;
```
FellowTraveler: Should put an OT_ASSERT_MSG above the return.

src/core/app/Dht.cpp
```diff
@@ -0,0 +1,279 @@
...
+}
+
+Dht& Dht::It(DhtConfig& config)
+{
+    if (nullptr == instance_)
+    {
+        instance_ = new Dht(config);
+    }
+
+    return *instance_;
```
FellowTraveler: Should have an OT_ASSERT_MSG before the return

src/core/app/Dht.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+void Dht::Insert(
+    __attribute__((unused)) const std::string ID,
+    __attribute__((unused)) const Contract& contract)
+{
+#ifdef OT_DHT
+    OT_ASSERT(nullptr != node_);
+
+    String data(contract);
+    node_->Insert(ID, std::string(data.Get()));
```
FellowTraveler: Can remove the std::string, unnecessary construction.

(Unless that 2nd parameter takes a move reference.)

src/core/app/Dht.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+        OTServerContract newContract;
+        bool loaded = newContract.LoadContractFromString(data);
+
+        if (!loaded) { continue; }
+
+        Identifier serverID;
+        newContract.SetIdentifier(contractID);
+
+        if (!newContract.VerifyContract()) { continue; }
```
FellowTraveler: First we load the contract from string, and make sure it loads properly.

Then we set its ID to the ID we are expecting it to have.

Then we call VerifyContract(), which calls VerifyID(), which compares the ID we were expecting, to the ID of the content we actually loaded.

Those are the same, meaning we got the contract we were expecting.

src/core/app/Dht.cpp
```diff
@@ -0,0 +1,290 @@
...
+}
+#endif
+
+void Dht::Cleanup()
+{
+#ifdef OT_DHT
+    delete node_;
+    node_ = nullptr;
+#endif
+}
```
FellowTraveler: Should be:

```
#ifdef OT_DHT
    if (nullptr != node_)
        delete node_;
    node_ = nullptr;
#endif
```

src/core/app/Settings.cpp
```diff
@@ -65,6 +65,29 @@ class Settings::SettingsPvt
...
+        if (!Save()) return false;
+    }
+
+    if (!Reset()) return false;
+
+    // Second Load, Throw Assert if Failed loading.
+    if (!Load()) {
+        otErr << __FUNCTION__
+              << ": Error: Unable to load config file."
+              << " It should exist, as we just saved it!\n";
```
FellowTraveler: We did not necessarily just save it.

We actually only saved it if the first load failed. So this log is inaccurate. And frankly, I find it rather tacky.

src/network/OpenDHT.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+OpenDHT& OpenDHT::It(DhtConfig& config)
+{
+    if (nullptr == instance_)
+    {
+        instance_ = new OpenDHT(config);
+    }
+
+    return *instance_;
```
FellowTraveler: Should have an OT_ASSERT_MSG here.

src/storage/Storage.cpp
```diff
@@ -50,33 +50,38 @@
...
 {
 
     if (nullptr == instance_pointer_) {
 #ifdef OT_STORAGE_FS
-        instance_pointer_ = new StorageFS(param, hash, random);
+        instance_pointer_ = new StorageFS(config, hash, random);
 #elif defined OT_STORAGE_SQLITE
-        instance_pointer_ = new StorageSqlite3(param, hash, random);
+        instance_pointer_ = new StorageSqlite3(config, hash, random);
 #endif
```
FellowTraveler:
```
#else
#error "blah blah blah"
#endif
```

PR #656, opened 2016-01-13
Implement verification credentials

justusranvier:
Verification credentials contain signed assertions regarding the validity of individual claims.

### DIFF
--- a/src/core/crypto/KeyCredential.cpp
+++ b/src/core/crypto/KeyCredential.cpp
@@ -211,13 +211,6 @@ int32_t KeyCredential::GetPublicKeysBySignature(
     return nCount;
 }
 
-// Verify that m_strNymID is the same as the hash of m_strSourceForNymID.
-// Child key verifies (master does not): NymID against master credential NymID, and master
-// credential ID against hash of master credential.
-// (How? Because MasterCredential overrides this function and does NOT call the
-// parent, whereas ChildKeyCredential does.)
-// Then verify the (self-signed) signature on *this.
-//
 bool KeyCredential::VerifyInternally() const
 {
     // Perform common Credential verifications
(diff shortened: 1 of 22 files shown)


src/core/crypto/KeyCredential.cpp
```diff
@@ -490,14 +481,14 @@ bool KeyCredential::Sign(
     const proto::SignatureRole role,
     proto::KeyRole key) const
 {
-    const OTAsymmetricKey* keyToUse;
+    const OTKeypair* keyToUse;
```
FellowTraveler: Make sure the default case sets keyToUser to nullptr

src/core/crypto/KeyCredential.cpp
```diff
@@ -507,7 +498,32 @@ bool KeyCredential::Sign(
...
-    return keyToUse->Sign(plaintext, sig, pPWData, exportPassword, credID, role);
+    return keyToUse->Sign(plaintext, credID, sig, pPWData, exportPassword, role);
+}
+
+bool KeyCredential::Verify(
+    const OTData& plaintext,
+    proto::Signature& sig,
+    proto::KeyRole key) const
+{
+    const OTKeypair* keyToUse;
```
FellowTraveler: Make sure the default case sets keyToUser to nullptr, if you can't set it to nullptr on construction.

src/core/crypto/VerificationCredential.cpp
```diff
@@ -0,0 +1,174 @@
...
+    role_ = proto::CREDROLE_VERIFY;
+    nym_id_ = parent.GetNymID();
+    master_id_ = parent.GetMasterCredID();
+    auto verificationSet = nymParameters.VerificationSet();
+    if (verificationSet) {
+        data_.reset(new proto::VerificationSet(*verificationSet));
+    }
+
+    Identifier childID;
+    CalculateAndSetContractID(childID);
```
FellowTraveler: This stuff needs to be moved to Factory function.

PR #659, opened 2016-01-14
Cleanup and minor fixes

justusranvier:


src/core/crypto/ChildKeyCredential.cpp
```diff
@@ -85,19 +84,10 @@ ChildKeyCredential::ChildKeyCredential(CredentialSet& other, const proto::Creden
 ChildKeyCredential::ChildKeyCredential(CredentialSet& other, const NymParameters& nymParameters)
     : ot_super(other, nymParameters, proto::CREDROLE_CHILDKEY)
 {
-    m_strContractType = "KEY CREDENTIAL";
     role_ = proto::CREDROLE_CHILDKEY;
 
     nym_id_ = other.GetNymID();
```
FellowTraveler: Are these member variables on ChildKeyCredential or a base class? If ChildKeyCredential should initialize them right after super like this:

```
: ot_super(other, nymParameters, proto::CREDROLE_CHILDKEY), role_(proto::CREDROLE_CHILDKEY), nym_id_(other.GetNymID()) etc
```

PR #690, opened 2016-05-11
Server improvements

justusranvier:
The IP address and port on which a server binds its listening socket is now controlled via the server's configuration file rather than the server contract. This allows a server to operate in situations where the externally-visible IP address(or domain name)/port combination does not match the local interface.

Server contract creation is now non-interactive. The first time a server is started it will create a nym and a contract using default values. Any or all of these defaults can be overridden via a string map passed into the opentxs::serverloader constructor.

Recognized keys:

terms: human-readable text which will be added to the contract (i.e. ToS)
externalip: the IP address or domain name to which the clients should connect
commandport: the TCP port to which clients should connect
bindip: the IP address to which the server should bind its listening socket
listencommand: the TCP port to which the server should bind its listening socket
name: human-readable server name

### DIFF
--- a/src/server/OTServer.cpp
+++ b/src/server/OTServer.cpp
@@ -224,7 +224,7 @@ void OTServer::CreateMainFile(
         terms = defaultTerms;
     }
 
-    const std::string defaultExternalIP = "127.0.0.1";
+    const std::string defaultExternalIP = DEFAULT_EXTERNAL_IP;
     const std::string& userExternalIP = args["externalip"];
     std::string hostname = userExternalIP;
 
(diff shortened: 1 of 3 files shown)


src/server/OTServer.cpp
```diff
@@ -781,11 +829,29 @@ bool OTServer::DropMessageToNymbox(const Identifier& NOTARY_ID,
 
 bool OTServer::GetConnectInfo(std::string& strHostname, uint32_t& nPort) const
 {
-    auto contract = App::Me().Contract().Server(m_strNotaryID);
+    bool notUsed;
```
FellowTraveler: should be =false

src/server/OTServer.cpp
```diff
@@ -781,11 +829,29 @@ bool OTServer::DropMessageToNymbox(const Identifier& NOTARY_ID,
...
+    int64_t port;
```
FellowTraveler: Should be =0; or =DEFAULT_IP

PR #733, opened 2016-07-22
Add ed25519 credentials

justusranvier:
* Add support for ed25519 keys and credentials
* Change default key credential type to ed25519
* Add support for multiple child key credentials
* Rename NewNymECDSA API calls and command line option to NewNymHD
* Regardless of which type of nym is created (legacy or HD), all supported child key credential types will be created
  * Legacy nyms have ed25519, secp256k1, and RSA credentials
  * HD nyms have ed25519 and secp256k1 credentials


src/core/Identifier.cpp
```diff
@@ -170,58 +175,85 @@ bool Identifier::operator>=(const Identifier& s2) const
...
-//
-void Identifier::GetString(String& theStr) const
+void Identifier::GetString(String& id) const
 {
-    App::Me().Crypto().Util().EncodeID(*this, theStr); // *this input, theStr output.
-}
+    OTData data;
+    data.Assign(&type_, sizeof(type_));
+
+    OT_ASSERT(1 == data.GetSize());
```
FellowTraveler: Fix all this.

Remove the block.

Make sure returning empty string on failure.

src/core/crypto/CryptoUtil.cpp
```diff
@@ -40,17 +40,43 @@
...
+
+    for (auto& character : input) {
+        output.push_back(character);
+
+        if (++width >= LineWidth) {
+            output.push_back('\n');
+            width = 0;
+        }
+    }
+    output.push_back('\n');
```
FellowTraveler: If width !=0 then do this push back (to avoid adding 2 newlines sometimes)

src/core/crypto/CryptoUtil.cpp
```diff
@@ -121,50 +147,90 @@ String CryptoUtil::Nonce(const uint32_t size, OTData& rawOutput) const
...
+        static_cast<const uint8_t*>(input.GetPointer()),
+        input.GetSize(),
+        breakLines);
+}
 
-    return String(::EncodeBase58Check(inputStart, inputEnd));
+std::string CryptoUtil::Base58CheckEncode(const OTPassword& input)
+{
+    return Base58CheckEncode(
+        static_cast<const uint8_t*>(input.getMemory()),
```
FellowTraveler: Need to check here for the password type (binary or text) and call the appropriate function. Cannot assume, it will assert if wrong.

src/core/crypto/CryptoUtil.cpp
```diff
@@ -121,50 +147,90 @@ String CryptoUtil::Nonce(const uint32_t size, OTData& rawOutput) const
...
+}
 
-        return false;
+bool CryptoUtil::Base58CheckDecode(const String& input, OTPassword& output)
+{
+    OTData decodedOutput;
+    bool decoded = Base58CheckDecode(input, decodedOutput);
+
+    if (decoded) {
+        output.setMemory(decodedOutput);
```
FellowTraveler: Need to check if binary or text type before calling this getMemory function. There will be a different one to call for text passwords.

If it's ALWAYS meant to be in binary mode, then need to assert that at the top of this function.

include/opentxs/core/crypto/TrezorCrypto.hpp
```diff
@@ -39,49 +39,117 @@
...
+        0xFF, 0xFF, 0xFF, 0xFF,
+        0xFF, 0xFF, 0xFF, 0xFF,
+        0xFF, 0xFF, 0xFF, 0xFF,
+        0xFF, 0xFF, 0xFF, 0xFF,
+        0xFF, 0xFF, 0xFF, 0xFF,
+        0xFF, 0xFF, 0xFF, 0xFE,
+        0xFF, 0xFF, 0xFC, 0x2F};
+
+#if OT_CRYPTO_WITH_BIP32
+    const curve_info* secp256k1_;
```
FellowTraveler: Need to initialize this to nullptr

src/core/NymIDSource.cpp
```diff
@@ -102,6 +113,29 @@ OTData NymIDSource::asData() const
...
+    const bool child = (proto::CREDROLE_CHILDKEY == credential.role());
+    const bool keyCredential = master || child;
+
+    if (!keyCredential) { return output; }
+
+    const auto& publicCred = credential.publiccredential();
+
+    for (auto& key : publicCred.key()) {
+        if (role == key.role()) {
+            output.reset(new proto::AsymmetricKey(key));
```
FellowTraveler: Need to break right here I think.

src/core/crypto/Ecdsa.cpp
```diff
@@ -0,0 +1,415 @@
...
+    OTData& encryptedKey)
+{
+    OTPassword keyPassword;
+    App::Me().Crypto().Hash().Digest(
+        proto::HASHTYPE_SHA256, password, keyPassword);
+
+    return App::Me().Crypto().AES().Encrypt(
+        CryptoSymmetric::AES_256_ECB,
+        keyPassword,
+        static_cast<const char*>(plaintextKey.getMemory()),
```
FellowTraveler: Need to verify that it's a binary password before calling getMemory or getMemorySize

Or at least assert, if it's supposed to be that way every time.

src/core/crypto/Libsecp256k1.cpp
```diff
@@ -60,19 +63,47 @@
...
+bool Libsecp256k1::RandomKeypair(
+    OTPassword& privateKey,
+    OTData& publicKey) const
+{
+    if (nullptr == context_) { return false; }
+
+    bool validPrivkey = false;
+    uint8_t candidateKey [PrivateKeySize]{};
+    uint8_t nullKey [PrivateKeySize]{};
+
```
FellowTraveler: int counter=0;

src/core/crypto/Libsecp256k1.cpp
```diff
@@ -60,19 +63,47 @@
...
+        // We add the random key to a zero value key because
+        // secp256k1_privkey_tweak_add checks the result to make sure it's in
+        // the correct range for secp256k1.
+        //
+        // This loop should almost always run exactly one time (about 1/(2^128)
+        // chance of randomly generating an invalid key thus requiring a second
+        // attempt)
+        validPrivkey = secp256k1_ec_privkey_tweak_add(
+            context_, candidateKey, nullKey);
+    }
```
FellowTraveler: (inside the while block)

assert(counter < 10);

src/core/crypto/Libsecp256k1.cpp
```diff
@@ -181,422 +213,85 @@ bool Libsecp256k1::OTDataToECDSASignature(
...
 
-        if (serialized) {
-            serializedPubkey.Assign(serializedOutput, serializedSize);
-            return serialized;
-        }
-    }
+    const auto created = secp256k1_ec_pubkey_create(
+        context_,
+        &key,
+        static_cast<const unsigned char*>(privateKey.getMemory()));
```
FellowTraveler: Should verify that the otpassword is in binary mode or it will assert here.

src/core/crypto/Libsodium.cpp
```diff
@@ -0,0 +1,287 @@
...
+}
+
+bool Libsodium::ECDH(
+    const OTData& publicKey,
+    const OTPassword& seed,
+    OTPassword& secret) const
+{
+    OTData notUsed;
+    OTPassword curvePrivate;
+
```
FellowTraveler: Should assert here that seed and secret are both binary (if they're meant to be.)

src/core/crypto/Libsodium.cpp
```diff
@@ -0,0 +1,287 @@
...
+        static_cast<const unsigned char*>(curvePublic.GetPointer()));
+
+    return (0 == output);
+}
+
+bool Libsodium::ExpandSeed(
+    const OTPassword& seed,
+    OTPassword& privateKey,
+    OTData& publicKey) const
+{
```
FellowTraveler: Should assert that privateKey is binary mode

PR #887, opened 2018-01-25
 Simplify the arguments to the Native factory methods. 

tomlandrum:


include/opentxs/OT.hpp
```diff
@@ -72,24 +93,13 @@ class OT
...
-        const bool recover,
-        const std::string& words,
-        const std::string& passphrase,
-        const std::chrono::seconds gcInterval = std::chrono::seconds(0),
-        const std::string& storagePlugin = "",
-        const std::string& backupDirectory = "",
-        const std::string& encryptedDirectory = "");
+        const bool recover = false);
     static void Join();
     static void ServerFactory(
```
justusranvier: For consistency the ServerFactory method should also take a "recover" argument.

include/opentxs/OT.hpp
```diff
@@ -40,12 +40,33 @@
...
 #include <chrono>
 #include <map>
 #include <string>
 
+#define OPENTXS_ARG_STORAGE_PLUGIN "storageplugin"
+#define OPENTXS_ARG_BACKUP_DIRECTORY "backupdirectory"
+#define OPENTXS_ARG_ENCRYPTED_DIRECTORY "encrypteddirectory"
+#define OPENTXS_ARG_WORDS "words"
+#define OPENTXS_ARG_PASSPHRASE "passphrase"
+
```
justusranvier: These two lists should be combined and sorted

include/opentxs/OT.hpp
FellowTraveler: /// recover - Regenerates the default BIP39 seed from the supplied words and passphrase.

PR #890, opened 2018-02-01
Added the original request ID to BailmentNotice.  Also added the vers…

tomlandrum:
…ion number to child classes of PeerRequest.

### DIFF
--- a/src/api/client/Pair.cpp
+++ b/src/api/client/Pair.cpp
@@ -439,6 +439,9 @@ void Pair::process_pending_bailment(
         issuer.AddRequest(proto::PEERREQUEST_PENDINGBAILMENT, requestID);
 
     if (added) {
+        const Identifier originalRequest(request.pendingbailment().requestid());
+        issuer.SetUsed(proto::PEERREQUEST_BAILMENT, originalRequest, true);
+
         OTAPI_Func operation(
             ACKNOWLEDGE_NOTICE,
             wallet_,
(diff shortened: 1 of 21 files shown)


src/core/contract/peer/PeerRequest.cpp
```diff
@@ -90,10 +90,11 @@ PeerRequest::PeerRequest(
 
 PeerRequest::PeerRequest(
     const ConstNym& nym,
+    const std::uint32_t version,
     const Identifier& recipient,
     const Identifier& server,
     const proto::PeerRequestType& type)
-    : ot_super(nym, PEER_REQUEST_VERSION)
```
justusranvier: Now that PEER_REQUEST_VERSION is no longer being used it should be removed from Types.hpp

src/api/client/Pair.cpp
```diff
@@ -439,6 +439,9 @@ void Pair::process_pending_bailment(
         issuer.AddRequest(proto::PEERREQUEST_PENDINGBAILMENT, requestID);
 
     if (added) {
+        const Identifier originalRequest(request.pendingbailment().requestid());
+        issuer.SetUsed(proto::PEERREQUEST_BAILMENT, originalRequest, true);
```
justusranvier: Should test originalRequest for empty() first before calling issuer.SetUsed

src/core/contract/peer/PeerRequest.cpp
tomlandrum: PEER_REQUEST_VERSION is still used in PeerObject and PeerReply. Do they also need to be changed to use CURRENT_VERSION?

justusranvier: All PeerReply objects should be created as version 4.

PeerObjects should be created as version 5.

PR #956, opened 2018-04-05
Added the NymFile interface and modified classes to use it.

tomlandrum:


include/opentxs/client/NymData.hpp
```diff
@@ -96,23 +98,26 @@ class NymData
         const bool primary,
         const bool active) const;
     std::string Name() const;
+    const class Nym& Nym() const;
```
justusranvier: This function also needs to be in the #ifndef SWIG section

src/api/client/Wallet.cpp
```diff
@@ -568,7 +551,30 @@ ConstNym Wallet::Nym(const proto::CredentialIndex& publicNym) const
...
+    auto pNym = std::make_unique<class Nym>(nymParameters);
+
+    if (pNym->VerifyPseudonym()) {
+        const bool nameAndTypeSet =
+            proto::CITEMTYPE_ERROR != type && !name.empty();
+        if (nameAndTypeSet) {
+            pNym->SetScope(type, name, true);
+
+            pNym->SetAlias(name);
+            pNym->SaveSignedNymfile(*pNym);
```
justusranvier: Saving the nymfile has nothing to do with setting the scope claim, so if it needs to be saved it should be done regardless.

PR #1019, opened 2018-05-16
Partial work on MultipartMessage classes and unit tests.

tomlandrum:


include/opentxs/network/zeromq/Message.hpp
```diff
@@ -75,7 +75,7 @@ class Message
     EXPORT virtual const void* data() const = 0;
     EXPORT virtual std::size_t size() const = 0;
 
-    EXPORT virtual operator zmq_msg_t*() = 0;
+    EXPORT virtual operator zmq_msg_t*() const = 0;
```
justusranvier: The reason is was not const before is that zmq_msg_send takes ownership of the object you pass in, so we want to make sure it's not possible to send a non-const Message.

src/network/zeromq/MultipartMessage.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+MultipartMessage* MultipartMessage::clone() const
+{
+    MultipartMessage* multipartMessage = new MultipartMessage();
+
+    OT_ASSERT(nullptr != multipartMessage);
+
+    for (auto& message : messages_) {
+        multipartMessage->messages_.push_back(message);
```
justusranvier: Wouldn't emplace_back avoid a copy here and other places where push_back is being used?

wrappers/opentxs.i
```diff
@@ -14,9 +14,12 @@
 #define SWIG_VERSION
 #endif
 
+#include "opentxs/network/zeromq/FrameIterator.hpp"
```
justusranvier: These includes should be added to <opentxs/opentxs.hpp> instead of being added to this file

src/network/zeromq/MultipartMessage.cpp
tomlandrum: Fixed.

wrappers/opentxs.i
tomlandrum: Fixed.

include/opentxs/network/zeromq/Message.hpp
tomlandrum: Fixed.

PR #1020, opened 2018-05-16
"const OTIdentifier" repair

Wallydog66:


include/opentxs/client/ServerAction.hpp
```diff
@@ -59,7 +59,7 @@ class ServerAction
 public:
     EXPORT virtual TransactionNumber GetTransactionNumber() const = 0;
     EXPORT virtual SendResult LastSendResult() const = 0;
-    EXPORT virtual const OTIdentifier MessageID() const = 0;
+    EXPORT virtual const Identifier& MessageID() const = 0;
```
justusranvier: This probably should be an OTIdentifier

PR #1031, opened 2018-05-22
Blockchain: Add tests for BIP32 M/0H/0/N addresses from seed

kmels:


src/api/Blockchain.cpp
```diff
@@ -459,7 +459,7 @@ void Blockchain::init_path(
     switch (standard) {
         case BlockchainAccountType::BIP32: {
             path.add_child(
-                account | static_cast<std::uint32_t>(Bip32Child::HARDENED));
+                0 | static_cast<std::uint32_t>(Bip32Child::HARDENED));
```
justusranvier: This workaround is only valid for the first nym in the wallet.

The right way to fix this is to make sure account is initialized with a correct value.

tests/blockchain/Test_AllocateAddress.cpp
```diff
@@ -84,12 +84,12 @@ TEST(Test_Blockchain, testAllocateAddress)
     .Blockchain().AllocateAddress(
                                   Identifier(Alice),
                                   Identifier(AliceAccountID),
-                                  label, BTC);
+                                  label, false);
```
justusranvier: Use EXTERNAL_CHAIN here instead of false

tests/blockchain/main.cpp
```diff
@@ -40,6 +40,7 @@
 #include "OTTestEnvironment.hpp"
 
 int main(int argc, char **argv) {
+  system("rm -r $HOME/.ot/");
```
justusranvier: This needs to be documented somewhere so that people know it will happen if they run the tests

kmels: Added a warning to the README for now.

PR #1038, opened 2018-05-25
Use OTIdentifier to instantiate Identifier objects

Wallydog66:
Moved the constructors back to public for this round of OTIdentifier changes.

### DIFF
--- a/src/client/OTAPI_Exec.cpp
+++ b/src/client/OTAPI_Exec.cpp
@@ -4108,16 +4108,16 @@ std::string OTAPI_Exec::ProposePaymentPlan(
     OT_VERIFY_MIN_BOUND(PAYMENT_PLAN_LENGTH, OT_TIME_ZERO);
     OT_VERIFY_MIN_BOUND(PAYMENT_PLAN_MAX_PAYMENTS, 0);
 
-    std::unique_ptr<Identifier> angelSenderAcctId;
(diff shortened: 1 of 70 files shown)


src/client/OTAPI_Exec.cpp
```diff
@@ -4094,7 +4097,7 @@ std::string OTAPI_Exec::ProposePaymentPlan(
     // NOTE: Making this optional for this step. (Since sender account may
     // not be known until the customer receives / confirms the payment plan.)
     //  OT_VERIFY_ID_STR(SENDER_ACCT_ID); // Optional parameter.
-    OT_VERIFY_ID_STR(RECIPIENT_NYM_ID);
```
justusranvier: This probably needs to go back in

src/client/OT_API.cpp
```diff
@@ -8026,12 +8028,7 @@ bool OT_API::AddBasketExchangeItem(
 
     auto account = wallet_.Account(ASSET_ACCOUNT_ID);
 
-    if (false == bool(account)) {
```
justusranvier: This error message should go back in

src/client/OT_API.cpp
```diff
@@ -9303,12 +9304,7 @@ bool OT_API::DiscardCheque(
 
     auto account = wallet_.Account(accountID);
 
-    if (false == bool(account)) {
```
justusranvier: Put error message back

PR #1043, opened 2018-06-13
Nym::Bip47 Store and Load 

kmels:


src/storage/tree/Nym.cpp
```diff
@@ -1095,11 +1126,54 @@ bool Nym::Store(
...
+
+        return false;
+    }
+
+    Lock writeLock(write_lock_, std::defer_lock);
+    sLock bip47Lock(bip47_lock_, std::defer_lock);
+    std::lock(writeLock, bip47Lock);
+
+    for (const auto& chain : context.chain()) {
+        const auto it = bip47_contexts_.find(chain.type());
```
justusranvier: auto& chain = bip47_contexts_[chain.type()]

That line will create the inner map if it didn't already exist

src/storage/tree/Nym.cpp
```diff
@@ -1095,11 +1126,54 @@ bool Nym::Store(
...
+
+        if (it == bip47_contexts_.end()) {
+            bip47_contexts_.emplace(
+                chain.type(),
+                std::map<std::string, std::set<Bip47ChannelID>>());
+        }
+
+        for (const auto& channel : chain.channel()) {
+            const auto innerMap = it->second;
```
justusranvier: auto& channels = chain[channel.contact()];
channels.emplace(Bip47ChannelID{channel.contact(), channel.paymentcode()});

src/storage/tree/Nym.cpp
kmels: great !! added.

src/storage/tree/Nym.cpp
kmels: Thank you, I changed it

src/storage/tree/Nym.cpp
```diff
@@ -1067,6 +1081,25 @@ proto::StorageNym Nym::serialize() const
         *serialized.add_blockchainaccount() = account;
     }
 
+    for (const auto& it_chain : bip47_channels_) {
+        auto& chain = it_chain.first;
+
+        for (const auto& it_contact : it_chain.second) {
+            auto& channel = *serialized.add_bip47channelindex();
+            channel.set_version(1);
```
justusranvier: Use a define here so that the version number can be updated later

PR #1169, opened 2018-09-04
 Use OTString to instantiate String objects: 06 

Wallydog66:
Passes all tests.

### DIFF
--- a/src/core/trade/OTMarket.cpp
+++ b/src/core/trade/OTMarket.cpp
@@ -122,16 +122,18 @@ std::int32_t OTMarket::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
     //    return nReturnVal;
 
     if (!strcmp("market", xml->getNodeName())) {
-        m_strVersion = xml->getAttributeValue("version");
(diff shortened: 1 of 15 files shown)


src/core/trade/OTMarket.cpp
```diff
@@ -1252,7 +1259,7 @@ void OTMarket::ProcessTrade(
                  "trader's accounts in OTMarket::"
               << __FUNCTION__ << "\n";
         theTrade.FlagForRemoval();  // Removes from Cron.
-        return;
```
FellowTraveler: Logic change?

src/core/trade/OTMarket.cpp
```diff
@@ -2650,13 +2661,13 @@ bool OTMarket::ValidateOfferForMarket(OTOffer& theOffer, String* pReason)
         otOut << __FUNCTION__
               << ": Offer is invalid for this market: " << strReason << "\n";
 
-        if (nullptr != pReason) *pReason = strReason;
+        if (!pReason->Get()) pReason = strReason;
```
FellowTraveler: Should be Exists not Get

PR #1170, opened 2018-09-05
Use OTString to instantiate String objects: OTSmartContract.hpp

Wallydog66:
Passes all tests.

### DIFF
--- a/src/core/script/OTSmartContract.cpp
+++ b/src/core/script/OTSmartContract.cpp
@@ -729,6 +729,10 @@ typedef bool (OTSmartContract::*OT_SM_RetBool_ThrStr)(
 OTSmartContract::OTSmartContract(const api::Core& core)
     : ot_super(core)
     , m_StashAccts(core, Account::stash)
+    , m_strLastSenderUser(String::Factory())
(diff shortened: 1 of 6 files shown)


src/core/script/OTSmartContract.cpp
```diff
@@ -1491,7 +1500,7 @@ std::string OTSmartContract::GetStashBalance(
     OT_ASSERT(nullptr != pCron);
 
     ConstNym pServerNym = pCron->GetServerNym();
-    OT_ASSERT(nullptr != pServerNym);
+    OT_ASSERT(!pServerNym);
```
FellowTraveler: Here's a bug.

src/core/script/OTSmartContract.cpp
```diff
@@ -3575,8 +3587,9 @@ void OTSmartContract::ExecuteClauses(
             } else  // The param_string variable isn't already there. (So we add
                     // it as blank, if a value wasn't passed in.)
             {
-                if (nullptr != pParam)  // if a param was passed in...
+                if (pParam->empty()) { // if a param was passed in...
```
FellowTraveler: this should be !empty()

PR #1173, opened 2018-09-06
Use OTString to instantiate String objects: Encode.hpp

Wallydog66:
Passes all tests.

### DIFF
--- a/include/opentxs/Proto.hpp
+++ b/include/opentxs/Proto.hpp
@@ -56,11 +56,11 @@ std::string ProtoAsString(const T& serialized)
 }
 
 template <class T>
-String ProtoAsArmored(const T& serialized, const String& header)
+OTString ProtoAsArmored(const T& serialized, const String& header)
 {
     auto data = ProtoAsData<T>(serialized);
     Armored armored(data);
-    String output;
+    auto output = String::Factory();
     armored.WriteArmoredString(output, header.Get());
 
     return output;
@@ -91,7 +91,7 @@ template <class T>
 T StringToProto(const String& input)
 {
     Armored armored;
-    String unconstInput(input);
+    OTString unconstInput = String::Factory(input.Get());
 
     if (!armored.LoadFromString(unconstInput)) {
         std::cerr << __FUNCTION__ << "Failed to decode armored protobuf."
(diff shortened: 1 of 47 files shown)


Wallydog66:
*** DOES NOT PASS TESTS YET ***

include/opentxs/Proto.hpp
```diff
@@ -91,7 +91,7 @@ template <class T>
 T StringToProto(const String& input)
 {
     Armored armored;
-    String unconstInput(input);
+    OTString unconstInput(input);
```
FellowTraveler: Shouldn't this be: 
   OTString unconstInput = String::Factory(input);

src/client/OTAPI_Exec.cpp
```diff
@@ -9893,7 +9936,7 @@ std::string OTAPI_Exec::Purse_Push(
                                                                      // purse is
     // password-protected, then this
     // parameter should be "".
-    strReason.empty();
+    strReason->empty();
```
FellowTraveler: This makes no sense.

strReason->empty(); <=== This just returns true/false based on whether or not the string is empty. Then we're at the end of the statement and that boolean value is simply discarded unused.

Probably meant to do strReason->clear();   ????  That will clear the string and MAKE it empty.

I know the bug was introduced in some previous PR (not this one) but it's definitely a bug.

src/client/OT_API.cpp
```diff
@@ -9663,8 +9697,8 @@ CommandResult OT_API::triggerClause(
 
     // Optional string parameter. Available as "param_string" inside the
     // script.
-    if ((nullptr != pStrParam) && (pStrParam->Exists())) {
-        payload.SetString(*pStrParam);
+    if ((pStrParam.Exists()) && (pStrParam.Exists())) {
```
FellowTraveler: Change this line:
    if ((pStrParam.Exists()) && (pStrParam.Exists())) {

To this:
   if (pStrParam.Exists()) {

(No need to repeat the same expression twice. The old code was checking the pointer for null, and then checking the string for Exists(). The new code is no longer using a raw pointer).

src/core/contract/ServerContract.cpp
```diff
@@ -201,9 +201,9 @@ proto::ServerContract ServerContract::IDVersion(const Lock& lock) const
     contract.clear_publicnym();  // reinforcing that this field must be blank.
 
     if (nullptr != nym_) {
```
FellowTraveler: Change:
   if (nullptr != nym_) {

To:
   if (nym_) {

Looks like this line never got fixed once the nym_ member variable was changed from a raw pointer to a smart pointer.

PR #1190, opened 2018-09-18
Use OTString to instantiate String objects: Message.hpp

Wallydog66:
Passes all tests.

### DIFF
--- a/src/client/OTClient.cpp
+++ b/src/client/OTClient.cpp
@@ -219,7 +219,7 @@ bool OTClient::AcceptEntireNymbox(
                      "replyNotice, either!)\n";
         }
 
-        String strRespTo;
+        auto strRespTo = String::Factory();
         pTransaction->GetReferenceString(strRespTo);
 
         if ((transactionType::message == pTransaction->GetType())) {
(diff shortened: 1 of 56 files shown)


src/client/OTClient.cpp
```diff
@@ -4858,7 +4871,7 @@ bool OTClient::processServerReplyProcessNymbox(
     OT_ASSERT(nullptr != pReplyTransaction);
 
     const auto& NYM_ID = context.Nym()->ID();
-    const String& strNotaryID = String(context.Server());
+    const String& strNotaryID = String::Factory(context.Server());
```
FellowTraveler: This is bad. It's a reference to a temporary.
Replace with:

OTString strNotaryID = String::Factory(context.Server());

src/client/OTClient.cpp
```diff
@@ -7296,9 +7310,9 @@ std::int32_t OTClient::ProcessUserCommand(
             auto NYMBOX_HASH = Identifier::Factory(context.LocalNymboxHash());
             NYMBOX_HASH->GetString(theMessage.m_strNymboxHash);
 
-            if (!String(NYMBOX_HASH).Exists()) {
+            if (!String::Factory(NYMBOX_HASH)->Exists()) {
```
FellowTraveler: Do this instead:

if (NYMBOX_HASH.IsEmpty())

src/server/PayDividendVisitor.cpp
```diff
@@ -283,7 +285,7 @@ bool PayDividendVisitor::Trigger(
 PayDividendVisitor::~PayDividendVisitor()
 {
 
-    if (nullptr != m_pstrMemo) delete m_pstrMemo;
+    if (m_pstrMemo->Exists()) delete m_pstrMemo->Get();
```
FellowTraveler: This is bad.

m_pstrMemo is a smart pointer now. That means it manages its own memory.

That means you CANNOT be deleting its internal pointer like this. 

(No more managing memory by hand like with raw pointers.)

This will cause a double delete crash, since the object will later destruct and meanwhile you already deleted its internal member variable (unbeknownst to it).

FellowTraveler: Basically that entire line should just be deleted.

PR #1195, opened 2018-09-25
Patch around unused variable errors.

nopdotcom:
Building with gcc-8 causes unused variable warnings, which are fatal because of -Werror.


justusranvier:
Does adding the [[maybe_unused]] attribute to the affected variables fix the problem without needing gcc-specific pragmas?

nopdotcom:
It probably does. C++17 is a little new for me. I'll rewrite this.

justusranvier:
Closing due to conflicts

PR #1203, opened 2018-09-27
Use OTString to instantiate String objects: OTTransaction.cpp

Wallydog66:
Passes all tests.

### DIFF
--- a/src/core/NymFile.cpp
+++ b/src/core/NymFile.cpp
@@ -93,8 +93,8 @@ NymFile::NymFile(
     , m_lUsageCredits(0)
     , m_bMarkForDeletion(false)
     , m_strNymFile(String::Factory())
-    , m_strVersion(NYMFILE_VERSION)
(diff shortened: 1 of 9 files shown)


src/core/NymFile.cpp
```diff
@@ -199,7 +200,7 @@ bool NymFile::deserialize_nymfile(
                             .Flush();
                     }
                     bSuccess = true;
-                    convert = (String("1.0") == m_strVersion);
+                    convert = (String::Factory("1.0")->Compare(m_strVersion));
```
FellowTraveler: ```convert = (String::Factory("1.0")->Compare(m_strVersion));```

^^^ The above works. But you could have done this:

```convert = m_strVersion->Compare("1.0");```

Or worst case:

```convert = m_strVersion->Compare(String::Factory("1.0"));```

Or even probably:

```convert = m_strVersion->Compare(std::string("1.0"));```

PR #1211, opened 2018-10-02
Convert Armored to use Factory() call: Proto.cpp

Wallydog66:

### DIFF
--- a/src/core/crypto/OTKeyring.cpp
+++ b/src/core/crypto/OTKeyring.cpp
@@ -184,7 +184,7 @@ bool OTKeyring::Windows_StoreSecret(
         otErr << __FUNCTION__
               << ": Error: Output of Win32 CryptProtectData was empty.\n";
     } else {
-        Armored ascData(theOutput);
+        auto ascData = Armored::Factory(theOutput);
         const OTString strFoldername("win32_data");  // todo hardcoding.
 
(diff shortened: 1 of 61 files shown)


src/core/crypto/OTKeyring.cpp
```diff
@@ -1301,8 +1301,10 @@ bool OTKeyring::FlatFile_DeleteSecret(
                   << str_ExactPath.c_str() << '\n';
         } else {
             bSuccess = true;
-            LogVerbose(OT_METHOD)(__FUNCTION__) (": ** (OTKeyring::FlatFile_DeleteSecret) Success ")
-                   ("deleting file: ") (str_ExactPath.c_str()).Flush();
+            LogVerbose(OT_METHOD)(__FUNCTION__)(
+                ": ** (OTKeyring::FlatFile_DeleteSecret) Success ")(
```
FellowTraveler: Remove the part that says ```(OTKeyring::FlatFile_DeleteSecret)```

(Because that is now already handled by ```(OT_METHOD)(__FUNCTION__)``` section).

FellowTraveler: Notice BTW, that is the exact name of the function where this log appears.

Wallydog66:
This is done.

PR #1222, opened 2018-10-08
Use OTArmored to instantiate Armored Objects: Token.hpp

Wallydog66:
Passes all tests.

### DIFF
--- a/src/cash/Token.cpp
+++ b/src/cash/Token.cpp
@@ -182,23 +182,17 @@ void Token::Release()
 void Token::ReleasePrototokens()
 {
     for (auto& it : m_mapPublic) {
-        Armored* pPrototoken = it.second;
+        auto& pPrototoken = it.second;
         OT_ASSERT_MSG(
-            nullptr != pPrototoken,
+            !pPrototoken->Exists(),
             "nullptr Armored pointer in Token::ReleasePrototokens.");
-
-        delete pPrototoken;
-        pPrototoken = nullptr;
     }
 
     for (auto& it : m_mapPrivate) {
-        Armored* pPrototoken = it.second;
+        auto& pPrototoken = it.second;
         OT_ASSERT_MSG(
-            nullptr != pPrototoken,
+            !pPrototoken->Exists(),
             "nullptr Armored pointer in Token::ReleasePrototokens.");
-
-        delete pPrototoken;
-        pPrototoken = nullptr;
     }
 
     m_mapPublic.clear();
@@ -515,8 +509,8 @@ void Token::UpdateContents()
         tagProtoPurse->add_attribute("chosenIndex", formatInt(m_nChosenIndex));
 
         for (auto& it : m_mapPublic) {
-            Armored* pPrototoken = it.second;
-            OT_ASSERT(nullptr != pPrototoken);
+            auto& pPrototoken = it.second;
+            OT_ASSERT(!pPrototoken->Exists());
             tagProtoPurse->add_tag("prototoken", pPrototoken->Get());
         }
 
@@ -529,8 +523,8 @@ void Token::UpdateContents()
         TagPtr tagPrivateProtoPurse(new Tag("privateProtopurse"));
 
         for (auto& it : m_mapPrivate) {
-            Armored* pPrototoken = it.second;
-            OT_ASSERT(nullptr != pPrototoken);
+            auto& pPrototoken = it.second;
+            OT_ASSERT(!pPrototoken->Exists());
             tagPrivateProtoPurse->add_tag(
                 "privatePrototoken", pPrototoken->Get());
         }
@@ -638,20 +632,16 @@ std::int32_t Token::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
 
         return 1;
     } else if (strNodeName->Compare("prototoken")) {
-        Armored* pArmoredPrototoken = new Armored;
-        OT_ASSERT(nullptr != pArmoredPrototoken);
+        auto pArmoredPrototoken = Armored::Factory();
 
-        if (!Contract::LoadEncodedTextField(xml, *pArmoredPrototoken) ||
+        if (!Contract::LoadEncodedTextField(xml, pArmoredPrototoken) ||
             !pArmoredPrototoken->Exists()) {
             otErr << "Error in Token::ProcessXMLNode: prototoken field "
                      "without value.\n";
 
-            delete pArmoredPrototoken;
-            pArmoredPrototoken = nullptr;
-
             return (-1);  // error condition
         } else {
-            m_mapPublic[nPublicTokenCount] = pArmoredPrototoken;
+            m_mapPublic.emplace(nPublicTokenCount, pArmoredPrototoken);
             nPublicTokenCount++;
         }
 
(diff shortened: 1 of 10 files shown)


src/cash/Token.cpp
```diff
@@ -638,20 +632,16 @@ std::int32_t Token::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
...
             otErr << "Error in Token::ProcessXMLNode: prototoken field "
                      "without value.\n";
 
-            delete pArmoredPrototoken;
-            pArmoredPrototoken = nullptr;
-
             return (-1);  // error condition
         } else {
-            m_mapPublic[nPublicTokenCount] = pArmoredPrototoken;
+            m_mapPublic.emplace(nPublicTokenCount, pArmoredPrototoken);
```
FellowTraveler: You have to pass the Factory call to the emplace call. You can't use a variable in between.

So it should be like this:
```m_mapPublic.emplace(nPublicTokenCount, Armored::Factory(pArmoredPrototoken));```

File: src/cash/Token.cpp [hunk 2]
```diff
@@ -515,8 +509,8 @@ void Token::UpdateContents()
         tagProtoPurse->add_attribute("chosenIndex", formatInt(m_nChosenIndex));
 
         for (auto& it : m_mapPublic) {
-            Armored* pPrototoken = it.second;
-            OT_ASSERT(nullptr != pPrototoken);
+            auto& pPrototoken = it.second;
+            OT_ASSERT(!pPrototoken->Exists());
```
FellowTraveler: need to assert that it DOES exist, before using it. Not the opposite.

src/cash/Token.cpp
```diff
@@ -529,8 +523,8 @@ void Token::UpdateContents()
         TagPtr tagPrivateProtoPurse(new Tag("privateProtopurse"));
 
         for (auto& it : m_mapPrivate) {
-            Armored* pPrototoken = it.second;
-            OT_ASSERT(nullptr != pPrototoken);
+            auto& pPrototoken = it.second;
+            OT_ASSERT(!pPrototoken->Exists());
```
FellowTraveler: need to assert that it DOES exist, before using it. Not the opposite.

src/cash/Token.cpp
```diff
@@ -661,20 +651,16 @@ std::int32_t Token::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
...
             otErr << "Error in Token::ProcessXMLNode: privatePrototoken "
                      "field without value.\n";
 
-            delete pArmoredPrototoken;
-            pArmoredPrototoken = nullptr;
-
             return (-1);  // error condition
         } else {
-            m_mapPrivate[nPrivateTokenCount] = pArmoredPrototoken;
+            m_mapPrivate.emplace(nPrivateTokenCount, pArmoredPrototoken);
```
FellowTraveler: You have to pass the Factory call to the emplace call. You can't use a variable in between.

So it should be like this:
```m_mapPrivate.emplace(nPrivateTokenCount, Armored::Factory(pArmoredPrototoken));```

File: src/cash/Token.cpp [hunk 5]
```diff
@@ -695,13 +681,13 @@ bool Token::GetPrototoken(Armored& ascPrototoken, std::int32_t nTokenIndex)
     if (nTokenIndex >= m_nTokenCount) { return false; }
 
     for (auto& it : m_mapPublic) {
-        Armored* pPrototoken = it.second;
-        OT_ASSERT(nullptr != pPrototoken);
+        auto& pPrototoken = it.second;
+        OT_ASSERT(!pPrototoken->Exists());
```
FellowTraveler: This can't be right.

src/cash/Token.cpp
```diff
@@ -718,13 +704,13 @@ bool Token::GetPrivatePrototoken(
     if (nTokenIndex >= m_nTokenCount) { return false; }
 
     for (auto& it : m_mapPrivate) {
-        Armored* pPrototoken = it.second;
-        OT_ASSERT(nullptr != pPrototoken);
+        auto& pPrototoken = it.second;
+        OT_ASSERT(!pPrototoken->Exists());
```
FellowTraveler: This can't be right.

src/cash/TokenLucre.cpp
```diff
@@ -201,10 +201,10 @@ bool Token_Lucre::GenerateTokenRequest(
             theEnvelope.Seal(theNym, strPrivateCoin);  // Todo check the return
                                                        // values on these two
                                                        // functions
-            theEnvelope.GetCiphertext(*pArmoredPrivate);
+            theEnvelope.GetCiphertext(pArmoredPrivate);
 
-            m_mapPublic[i] = pArmoredPublic;
-            m_mapPrivate[i] = pArmoredPrivate;
+            m_mapPublic.emplace(i, pArmoredPublic);
```
FellowTraveler: ```m_mapPublic.emplace(i, Armored::Factory(pArmoredPublic));```

src/cash/TokenLucre.cpp
```diff
@@ -201,10 +201,10 @@ bool Token_Lucre::GenerateTokenRequest(
...
+            m_mapPrivate.emplace(i, pArmoredPrivate);
```
FellowTraveler: ```m_mapPrivate.emplace(i, Armored::Factory(pArmoredPrivate));```

src/core/NymIDSource.cpp
```diff
@@ -280,9 +280,9 @@ bool NymIDSource::Sign(
 OTString NymIDSource::asString() const
 {
     auto dataSource = asData();
-    Armored armoredSource(dataSource);
+    auto armoredSource = Armored::Factory(dataSource);
 
-    return armoredSource;
+    return armoredSource.get();
```
FellowTraveler: ```return armoredSource.get();```

Should be:

```return String::Factory(armoredSource->Get());```

Because it returns an OTString. The reason for the ->Get() is so the String contains a copy of the armored string, rather than a decoded version of the armored string (which would be bad).

File: src/crypto/key/RSA.cpp [hunk 10]
```diff
@@ -139,22 +139,22 @@ RSA* RSA::clone() const
 
     OT_ASSERT(nullptr != key)
 
-    if (nullptr != m_p_ascKey) {
-        if (nullptr != key->m_p_ascKey) { delete key->m_p_ascKey; }
```
FellowTraveler: ```if (nullptr != key->m_p_ascKey)```

This says, IF THE KEY IS NOT NULL

src/crypto/key/RSA.cpp
```diff
@@ -139,22 +139,22 @@ RSA* RSA::clone() const
...
+    if (!m_p_ascKey->Exists()) {
+        if (!key->m_p_ascKey->Exists()) { key->m_p_ascKey->empty(); }
```
FellowTraveler: ```if (!m_p_ascKey->Exists())```

This says, IF THE KEY DOES NOT EXIST

src/crypto/key/RSA.cpp
FellowTraveler: It says it in two places actually. On the above two lines of code.

src/crypto/key/RSA.cpp
FellowTraveler: It says it in two places actually.

Haven't you changed the original logic?

src/crypto/key/RSA_private.cpp
```diff
@@ -123,10 +123,10 @@ EVP_PKEY* RSA::d::GetKeyLowLevel() const { return m_pKey; }
 const EVP_PKEY* RSA::d::GetKey(const OTPasswordData* pPWData)
 {
     OT_ASSERT_MSG(
-        nullptr != backlink->m_p_ascKey,
+        !backlink->m_p_ascKey->Exists(),
         "RSA::GetKey: nullptr != m_p_ascKey\n");
```
FellowTraveler: Above you changed: IF BACKLINK'S KEY IS NOT NULL

Into: IF BACKLINK'S KEY DOES NOT EXIST

src/crypto/key/RSA_private.cpp
```diff
@@ -123,10 +123,10 @@ EVP_PKEY* RSA::d::GetKeyLowLevel() const { return m_pKey; }
...
 
-    if (nullptr == backlink->m_p_ascKey) {
+    if (backlink->m_p_ascKey->Exists()) {
         otErr << __FUNCTION__
```
FellowTraveler: You changed: IF BACKLINK'S KEY IS NULL

Into:  IF BACKLINK'S KEY EXISTS

PR #1223, opened 2018-10-09
Use OTArmored to instantiate Armored Objects: Mint.cpp

Wallydog66:


src/cash/Mint.cpp
```diff
@@ -134,16 +134,16 @@ bool Mint::Expired() const
 void Mint::ReleaseDenominations()
 {
     while (!m_mapPublic.empty()) {
-        Armored* pArmor = m_mapPublic.begin()->second;
+        const auto& pArmor = m_mapPublic.begin()->second;
         m_mapPublic.erase(m_mapPublic.begin());
-        delete pArmor;
-        pArmor = nullptr;
+        pArmor->Get();
+
```
FellowTraveler: You need to delete ALL the shit in this function and replace with:

m_mapPublic.clear();
m_mapPrivate.clear();

src/cash/Mint.cpp
```diff
@@ -409,12 +409,12 @@ bool Mint::VerifyContractID() const
 bool Mint::GetPrivate(Armored& theArmor, std::int64_t lDenomination)
 {
     for (auto& it : m_mapPrivate) {
-        Armored* pArmor = it.second;
+        const auto& pArmor = it.second;
         OT_ASSERT_MSG(
-            nullptr != pArmor, "nullptr mint pointer in Mint::GetPrivate.\n");
+            !pArmor->Exists(), "nullptr mint pointer in Mint::GetPrivate.\n");
```
FellowTraveler: You reversed the logic here...

src/cash/Mint.cpp
```diff
@@ -426,12 +426,12 @@ bool Mint::GetPrivate(Armored& theArmor, std::int64_t lDenomination)
 bool Mint::GetPublic(Armored& theArmor, std::int64_t lDenomination)
 {
     for (auto& it : m_mapPublic) {
-        Armored* pArmor = it.second;
+        const auto& pArmor = it.second;
         OT_ASSERT_MSG(
-            nullptr != pArmor, "nullptr mint pointer in Mint::GetPublic.\n");
+            !pArmor->Exists(), "nullptr mint pointer in Mint::GetPublic.\n");
```
FellowTraveler: You reversed the logic here.

src/cash/Mint.cpp
```diff
@@ -469,9 +469,9 @@ std::int64_t Mint::GetDenomination(std::int32_t nIndex)
 
     for (auto it = m_mapPublic.begin(); it != m_mapPublic.end();
          ++it, nIterateIndex++) {
-        Armored* pArmor = it->second;
+        const auto& pArmor = it->second;
         OT_ASSERT_MSG(
-            nullptr != pArmor,
+            !pArmor->Exists(),
```
FellowTraveler: You reversed the logic here.

src/cash/Mint.cpp
```diff
@@ -513,9 +513,9 @@ void Mint::UpdateContents()
                                          // SetSavePrivateKeys() to set it true.
 
             for (auto& it : m_mapPrivate) {
-                Armored* pArmor = it.second;
+                const auto& pArmor = it.second;
                 OT_ASSERT_MSG(
-                    nullptr != pArmor,
+                    !pArmor->Exists(),
```
FellowTraveler: You reversed the logic here...

src/cash/Mint.cpp
```diff
@@ -528,9 +528,9 @@ void Mint::UpdateContents()
             }
         }
         for (auto& it : m_mapPublic) {
-            Armored* pArmor = it.second;
+            const auto& pArmor = it.second;
             OT_ASSERT_MSG(
-                nullptr != pArmor,
+                !pArmor->Exists(),
```
FellowTraveler: You reversed the logic here.

src/cash/Mint.cpp
```diff
@@ -612,43 +612,41 @@ std::int32_t Mint::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
         std::int64_t lDenomination =
             String::StringToLong(xml->getAttributeValue("denomination"));
 
-        Armored* pArmor = new Armored;
+        auto pArmor = Armored::Factory();
 
-        OT_ASSERT(nullptr != pArmor);
+        OT_ASSERT(!pArmor->Exists());
```
FellowTraveler: You reversed the logic here.

Anyway you need to delete this line entirely since it's now a smart pointer. WHY?

Because A raw pointer will be not-null even if the string itself is empty. 
Whereas "Exists" is only telling you whether the string is empty.

Therefore you need to delete the assert entirely.

src/cash/Mint.cpp
```diff
@@ -612,43 +612,41 @@ std::int32_t Mint::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
...
 
-        if (!Contract::LoadEncodedTextField(xml, *pArmor) ||
+        if (!Contract::LoadEncodedTextField(xml, pArmor) ||
             !pArmor->Exists()) {
             otErr << "Error in Mint::ProcessXMLNode: mintPrivateInfo field "
                      "without value.\n";
 
-            delete pArmor;
-            pArmor = nullptr;
+            pArmor->Get();
```
FellowTraveler: Why are you getting the raw pointer address here and then discarding it?

src/cash/Mint.cpp
```diff
@@ -612,43 +612,41 @@ std::int32_t Mint::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
...
 
             return (-1);  // error condition
         } else {
-            m_mapPrivate[lDenomination] = pArmor;
+            m_mapPrivate.emplace(lDenomination, pArmor);
```
FellowTraveler: You have to pass the Factory call into emplace.

src/cash/Mint.cpp
```diff
@@ -612,43 +612,41 @@ std::int32_t Mint::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
...
         return 1;
     } else if (strNodeName->Compare("mintPublicInfo")) {
         std::int64_t lDenomination =
             String::StringToLong(xml->getAttributeValue("denomination"));
 
-        Armored* pArmor = new Armored;
+        auto pArmor = Armored::Factory();
 
-        OT_ASSERT(nullptr != pArmor);
+        OT_ASSERT(!pArmor->Exists());
```
FellowTraveler: Remove this ASSERT line entirely. There is no longer a raw pointer involved.

src/cash/Mint.cpp
```diff
@@ -612,43 +612,41 @@ std::int32_t Mint::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
...
 
-        if (!Contract::LoadEncodedTextField(xml, *pArmor) ||
+        if (!Contract::LoadEncodedTextField(xml, pArmor) ||
             !pArmor->Exists()) {
             otErr << "Error in Mint::ProcessXMLNode: mintPublicInfo field "
                      "without value.\n";
 
-            delete pArmor;
-            pArmor = nullptr;
+            pArmor->Get();
```
FellowTraveler: Why are you calling Get() here to retrieve the raw pointer, and then discarding the result?

src/cash/Mint.cpp
```diff
@@ -612,43 +612,41 @@ std::int32_t Mint::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
...
 
             return (-1);  // error condition
         } else {
-            m_mapPublic[lDenomination] = pArmor;
+            m_mapPublic.emplace(lDenomination, pArmor);
```
FellowTraveler: You need to pass the Factory() call into emplace.

src/cash/MintLucre.cpp
```diff
@@ -133,11 +133,11 @@ bool MintLucre::AddDenomination(
...
 
-        Armored* pPublic = new Armored;
-        Armored* pPrivate = new Armored;
+        auto pPublic = Armored::Factory();
+        auto pPrivate = Armored::Factory();
 
-        OT_ASSERT(nullptr != pPublic);
-        OT_ASSERT(nullptr != pPrivate);
+        OT_ASSERT(!pPublic->Exists());
+        OT_ASSERT(!pPrivate->Exists());
```
FellowTraveler: You reversed the logic.

Anyway, in this case you want to delete the ASSERTs entirely.

src/cash/MintLucre.cpp
```diff
@@ -148,11 +148,11 @@ bool MintLucre::AddDenomination(
...
                                                       // values on these two
                                                       // functions
-        theEnvelope.GetCiphertext(*pPrivate);
+        theEnvelope.GetCiphertext(pPrivate);
 
         // Add the new key pair to the maps, using denomination as the key
-        m_mapPublic[lDenomination] = pPublic;
-        m_mapPrivate[lDenomination] = pPrivate;
+        m_mapPublic.emplace(lDenomination, pPublic);
+        m_mapPrivate.emplace(lDenomination, pPrivate);
```
FellowTraveler: You can't pass a variable into emplace.

You have to call the Factory and pass the result DIRECTLY into emplace. 

Meaning the MAP **IS** THE VARIABLE.

src/cash/Purse.cpp
```diff
@@ -1080,15 +1080,14 @@ std::int32_t Purse::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
 
         return 1;
     } else if (strNodeName->Compare("token")) {
-        Armored* pArmor = new Armored;
-        OT_ASSERT(nullptr != pArmor);
+        auto pArmor = Armored::Factory();
+        OT_ASSERT(!pArmor->Exists());
```
FellowTraveler: If you just created pArmor with a Factory call, then why are you assert that it does NOT exist??

Is that what you expect to be the case, honestly?

src/cash/Purse.cpp
```diff
@@ -1080,15 +1080,14 @@ std::int32_t Purse::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
...
 
-        if (!Contract::LoadEncodedTextField(xml, *pArmor) ||
+        if (!Contract::LoadEncodedTextField(xml, pArmor) ||
             !pArmor->Exists()) {
             otErr << szFunc << ": Error: token field without value.\n";
 
-            delete pArmor;
-            pArmor = nullptr;
+            pArmor->Get();
```
FellowTraveler: Why are you retrieving the raw pointer using Get(), and then discarding it?

src/cash/Purse.cpp
```diff
@@ -1213,10 +1212,9 @@ Token* Purse::Pop(OTNym_or_SymmetricKey theOwner)
     // Grab a pointer to the ascii-armored token, and remove it from the deque.
     // (And delete it.)
     //
-    Armored* pArmor = m_dequeTokens.front();
+    const auto& pArmor = m_dequeTokens.front();
     m_dequeTokens.pop_front();
-    delete pArmor;
-    pArmor = nullptr;
+    pArmor->Get();
```
FellowTraveler: Why are you retrieving the raw pointer here, and then discarding it, and not even using it for anything???

src/cash/Purse.cpp
```diff
@@ -1250,10 +1248,10 @@ void Purse::RecalculateExpirationDates(OTNym_or_SymmetricKey& theOwner)
     m_tEarliestValidTo = OT_TIME_ZERO;
 
     for (auto& it : m_dequeTokens) {
-        Armored* pArmor = it;
-        OT_ASSERT(nullptr != pArmor);
+        const auto& pArmor = it;
+        OT_ASSERT(!pArmor->Exists());
```
FellowTraveler: You should be asserting here that it DOES exist (since you are about to use it...)

Since normally we EXPECT it to exist, your assert should NORMALLY crash the application whenever tested.

src/cash/Purse.cpp
```diff
@@ -1316,7 +1314,7 @@ bool Purse::Push(OTNym_or_SymmetricKey theOwner, const Token& theToken)
             theOwner.Seal_or_Encrypt(theEnvelope, strToken, strDisplay);
 
         if (bSuccess) {
-            Armored* pArmor = new Armored(theEnvelope);
+            const auto& pArmor = Armored::Factory(theEnvelope);
```
FellowTraveler: reference to a temporary

src/cash/Purse.cpp
```diff
@@ -1382,9 +1380,9 @@ bool Purse::IsEmpty() const { return m_dequeTokens.empty(); }
 void Purse::ReleaseTokens()
 {
     while (!m_dequeTokens.empty()) {
-        Armored* pArmor = m_dequeTokens.front();
+        const auto& pArmor = m_dequeTokens.front();
         m_dequeTokens.pop_front();
-        delete pArmor;
+        pArmor->empty();
```
FellowTraveler: Why are you calling empty() (which returns true/false if pArmor is empty) and then discarding the result?

FURTHERMORE notice you are doing it right after the dequeTokens has called pop_front().

That means you got a reference to it, then you popped it off the deque (destroying it) and then you tried to access the destroyed object by dereferencing it with the empty() call, which will presumably crash the application. 

...All this when you simply discarded the result of empty(), after calling it.

src/cash/Token.cpp
```diff
@@ -182,23 +182,17 @@ void Token::Release()
 void Token::ReleasePrototokens()
 {
     for (auto& it : m_mapPublic) {
-        Armored* pPrototoken = it.second;
+        auto& pPrototoken = it.second;
         OT_ASSERT_MSG(
-            nullptr != pPrototoken,
+            !pPrototoken->Exists(),
```
FellowTraveler: You changed IF NOT NULL into IF DOESN'T EXIST

src/cash/Token.cpp
```diff
@@ -182,23 +182,17 @@ void Token::Release()
...
-        delete pPrototoken;
-        pPrototoken = nullptr;
     }
 
     for (auto& it : m_mapPrivate) {
-        Armored* pPrototoken = it.second;
+        auto& pPrototoken = it.second;
         OT_ASSERT_MSG(
-            nullptr != pPrototoken,
+            !pPrototoken->Exists(),
```
FellowTraveler: You changed IF NOT NULL into IF DOESN'T EXIST

Are you reading these out loud yet? Read them out loud.

src/cash/Token.cpp
```diff
@@ -515,8 +509,8 @@ void Token::UpdateContents()
         tagProtoPurse->add_attribute("chosenIndex", formatInt(m_nChosenIndex));
 
         for (auto& it : m_mapPublic) {
-            Armored* pPrototoken = it.second;
-            OT_ASSERT(nullptr != pPrototoken);
+            auto& pPrototoken = it.second;
+            OT_ASSERT(!pPrototoken->Exists());
             tagProtoPurse->add_tag("prototoken", pPrototoken->Get());
```
FellowTraveler: You changed IF NOT NULL into IF DOESN'T EXIST

src/cash/Token.cpp
```diff
@@ -529,8 +523,8 @@ void Token::UpdateContents()
         TagPtr tagPrivateProtoPurse(new Tag("privateProtopurse"));
 
         for (auto& it : m_mapPrivate) {
-            Armored* pPrototoken = it.second;
-            OT_ASSERT(nullptr != pPrototoken);
+            auto& pPrototoken = it.second;
+            OT_ASSERT(!pPrototoken->Exists());
             tagPrivateProtoPurse->add_tag(
```
FellowTraveler: You changed IF NOT NULL into IF DOESN'T EXIST

src/cash/Token.cpp
```diff
@@ -638,20 +632,16 @@ std::int32_t Token::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
...
             otErr << "Error in Token::ProcessXMLNode: prototoken field "
                      "without value.\n";
 
-            delete pArmoredPrototoken;
-            pArmoredPrototoken = nullptr;
-
             return (-1);  // error condition
         } else {
-            m_mapPublic[nPublicTokenCount] = pArmoredPrototoken;
+            m_mapPublic.emplace(nPublicTokenCount, pArmoredPrototoken);
```
FellowTraveler: You have to pass the Factory call into emplace.

src/cash/Token.cpp
```diff
@@ -661,20 +651,16 @@ std::int32_t Token::ProcessXMLNode(irr::io::IrrXMLReader*& xml)
...
             otErr << "Error in Token::ProcessXMLNode: privatePrototoken "
                      "field without value.\n";
 
-            delete pArmoredPrototoken;
-            pArmoredPrototoken = nullptr;
-
             return (-1);  // error condition
         } else {
-            m_mapPrivate[nPrivateTokenCount] = pArmoredPrototoken;
+            m_mapPrivate.emplace(nPrivateTokenCount, pArmoredPrototoken);
```
FellowTraveler: You have to pass the Factory call into emplace

src/cash/Token.cpp
```diff
@@ -695,13 +681,13 @@ bool Token::GetPrototoken(Armored& ascPrototoken, std::int32_t nTokenIndex)
     if (nTokenIndex >= m_nTokenCount) { return false; }
 
     for (auto& it : m_mapPublic) {
-        Armored* pPrototoken = it.second;
-        OT_ASSERT(nullptr != pPrototoken);
+        auto& pPrototoken = it.second;
+        OT_ASSERT(!pPrototoken->Exists());
```
FellowTraveler: You changed IF NOT NULL to IF DOESN'T EXIST

Do you read these out loud when you code them?

src/cash/Token.cpp
```diff
@@ -718,13 +704,13 @@ bool Token::GetPrivatePrototoken(
     if (nTokenIndex >= m_nTokenCount) { return false; }
 
     for (auto& it : m_mapPrivate) {
-        Armored* pPrototoken = it.second;
-        OT_ASSERT(nullptr != pPrototoken);
+        auto& pPrototoken = it.second;
+        OT_ASSERT(!pPrototoken->Exists());
```
FellowTraveler: You changed IF NOT NULL to IF DOESN'T EXIST

**There are only two possible answers: True or False.**

**How can you get it wrong more than 50% of the time??? Rolling dice would get it right more often.**

src/cash/TokenLucre.cpp
```diff
@@ -183,11 +183,11 @@ bool Token_Lucre::GenerateTokenRequest(
...
 
-            Armored* pArmoredPublic = new Armored(strPublicCoin);
-            Armored* pArmoredPrivate = new Armored;
+            auto pArmoredPublic = Armored::Factory(strPublicCoin);
+            auto pArmoredPrivate = Armored::Factory();
 
             OT_ASSERT_MSG(
-                ((nullptr != pArmoredPublic) && (nullptr != pArmoredPrivate)),
+                ((pArmoredPublic->Exists()) && (pArmoredPrivate->Exists())),
                 "ERROR: Unable to allocate memory in "
```
FellowTraveler: ^^^ **THIS ONE IS RIGHT!!!**

FellowTraveler: You did this one correctly!!!!!!!!!!!!!!

src/cash/TokenLucre.cpp
```diff
@@ -201,10 +201,10 @@ bool Token_Lucre::GenerateTokenRequest(
             theEnvelope.Seal(theNym, strPrivateCoin);  // Todo check the return
                                                        // values on these two
                                                        // functions
-            theEnvelope.GetCiphertext(*pArmoredPrivate);
+            theEnvelope.GetCiphertext(pArmoredPrivate);
 
-            m_mapPublic[i] = pArmoredPublic;
-            m_mapPrivate[i] = pArmoredPrivate;
+            m_mapPublic.emplace(i, pArmoredPublic);
+            m_mapPrivate.emplace(i, pArmoredPrivate);
```
FellowTraveler: You have to pass Factory into the above two emplace calls.

src/core/NymIDSource.cpp
```diff
@@ -280,9 +280,9 @@ bool NymIDSource::Sign(
 OTString NymIDSource::asString() const
 {
     auto dataSource = asData();
-    Armored armoredSource(dataSource);
+    auto armoredSource = Armored::Factory(dataSource);
 
-    return armoredSource;
+    return armoredSource.get();
```
FellowTraveler: Should be:

```return String::Factory(armoredSource->Get());```

File: src/crypto/key/RSA.cpp [hunk 32]
```diff
@@ -139,22 +139,22 @@ RSA* RSA::clone() const
 
     OT_ASSERT(nullptr != key)
 
-    if (nullptr != m_p_ascKey) {
-        if (nullptr != key->m_p_ascKey) { delete key->m_p_ascKey; }
+    if (!m_p_ascKey->Exists()) {
+        if (!key->m_p_ascKey->Exists()) { key->m_p_ascKey->empty(); }
```
FellowTraveler: Above, in multiple places, you changed IF NOT NULL into IF DOESN'T EXIST

FellowTraveler: Why are you calling empty, which returns true/false, and then discarding the result?

You probably meant to call clear() or Release()

src/crypto/key/RSA.cpp
```diff
@@ -139,22 +139,22 @@ RSA* RSA::clone() const
...
-        key->m_p_ascKey = new Armored(*m_p_ascKey);
+        key->m_p_ascKey = Armored::Factory(m_p_ascKey);
     }
 
     return key;
 }
 
 // virtual
-bool RSA::IsEmpty() const { return (nullptr == m_p_ascKey); }
+bool RSA::IsEmpty() const { return (m_p_ascKey->Get()); }
```
FellowTraveler: The old version returned TRUE or FALSE. **(Notice the return type of the function is bool).**

Whereas, _you_ are returning a RAW POINTER ADDRESS. Why?

You should probably return !m_p_ascKey->Exists();

src/crypto/key/RSA.cpp
```diff
@@ -139,22 +139,22 @@ RSA* RSA::clone() const
...
 
 // virtual
 bool RSA::GetPublicKey(String& strKey) const
 {
-    if (nullptr != m_p_ascKey) {
+    if (!m_p_ascKey->Exists()) {
         strKey.Concatenate(
```
FellowTraveler: You changed IF NOT NULL into IF DOESN'T EXIST

src/crypto/key/RSA.cpp
```diff
@@ -175,9 +175,9 @@ bool RSA::SetPublicKey(const String& strKey)
     m_bIsPublicKey = true;
     m_bIsPrivateKey = false;
 
-    if (nullptr == m_p_ascKey) {
-        m_p_ascKey = new Armored;
-        OT_ASSERT(nullptr != m_p_ascKey);
+    if (m_p_ascKey->Exists()) {
+        m_p_ascKey = Armored::Factory();
+        OT_ASSERT(m_p_ascKey->Exists());
     }
```
FellowTraveler: 1. You changed IF THE KEY IS NULL into IF THE KEY EXISTS.

FellowTraveler: 2. The old version ASSERTED that the raw pointer was NOT NULL. (Because it had just allocated the memory using a call to _new)_
Whereas the new version is ASSERTING that the string EXISTS (NOT EMPTY) -- of course that is a different meaning. You basically want to DELETE this assert. It no longer makes sense, since this is no longer a raw pointer.

src/crypto/key/RSA.cpp
```diff
@@ -601,7 +601,7 @@ bool RSA::ReEncryptPrivateKey(
                     // This base64 encodes the private key data, which
                     // is already encrypted to its passphase as well.
                     //
-                    m_p_ascKey->SetData(theNewData);  // <======== Success!
+                    Armored::Factory(m_p_ascKey)->SetData(theNewData);  // <======== Success!
                     bReturnVal = true;
```
FellowTraveler: The original code was already correct:  m_p_ascKey->SetData(theNewData);

You should change it back to the original code in this case.

src/crypto/key/RSA.cpp
```diff
@@ -776,7 +776,7 @@ std::shared_ptr<proto::AsymmetricKey> RSA::Serialize() const
     std::shared_ptr<proto::AsymmetricKey> serializedKey = ot_super::Serialize();
 
     auto dataKey = Data::Factory();
-    OT_ASSERT(m_p_ascKey);
+    OT_ASSERT(m_p_ascKey->Get());
     m_p_ascKey->GetData(dataKey);
```
FellowTraveler: You probably meant to assert that the Key "Exists."

You probably did NOT mean to assert that the Key's raw pointer address, when interpreted as though it were a true/false, evaluates to "true".

src/crypto/key/RSA.cpp
```diff
@@ -795,7 +795,7 @@ bool RSA::TransportKey(
     [[maybe_unused]] OTPassword& privateKey) const
 {
 #if OT_CRYPTO_SUPPORTED_KEY_ED25519
-    OT_ASSERT(nullptr != m_p_ascKey);
+    OT_ASSERT(!m_p_ascKey->Exists());
```
FellowTraveler: Here you reversed the logic. You changed IF NOT NULL into IF DOESN'T EXIST

src/crypto/key/RSA.cpp
```diff
@@ -820,8 +820,8 @@ RSA::~RSA()
     // Release the ascii-armored version of the key (safe to store in this
     // form.)
     //
-    if (nullptr != m_p_ascKey) delete m_p_ascKey;
-    m_p_ascKey = nullptr;
+    if (m_p_ascKey->Exists()) m_p_ascKey->empty();
```
FellowTraveler: 1. The Exists() call is CORRECT - good job! IF NOT NULL is the same as IF DOES EXIST

FellowTraveler: 2. However, the call to empty (which will ALWAYS return false, if the key DOES exist) is superfluous. It certainly doesn't DO anything, except return true/false to tell if you if the key is empty. You are simply **discarding the result**

Presumably you meant to call clear() or Release()

src/crypto/key/RSA.cpp
```diff
@@ -820,8 +820,8 @@ RSA::~RSA()
...
+    m_p_ascKey = Armored::Factory();
```
FellowTraveler: This call can just be deleted. 

The original code was just setting the raw pointer back to null after deleting it. 

But we aren't using raw pointers anymore. We're using a smart pointer.

src/crypto/key/RSA_private.cpp
```diff
@@ -91,9 +91,9 @@ void RSA::d::SetKeyAsCopyOf(
     backlink->m_bIsPublicKey = !bIsPrivateKey;
     backlink->m_bIsPrivateKey = bIsPrivateKey;
 
-    if (nullptr == backlink->m_p_ascKey) {
-        backlink->m_p_ascKey = new Armored;
-        OT_ASSERT(nullptr != backlink->m_p_ascKey);
+    if (backlink->m_p_ascKey->Exists()) {
```
FellowTraveler: You changed IF KEY IS NULL into IF KEY DOES EXIST.

src/crypto/key/RSA_private.cpp
```diff
@@ -91,9 +91,9 @@ void RSA::d::SetKeyAsCopyOf(
...
+        backlink->m_p_ascKey = Armored::Factory();
+        OT_ASSERT(!backlink->m_p_ascKey->Exists());
```
FellowTraveler: Here you created the backlink's key, then you assert that it does NOT exist. (Causing it to crash if it DOES exist).

So your logic is backwards. Anyway, you just want to delete this assert since it's no longer a raw pointer. You no longer need to assert that it has a valid raw address.

src/crypto/key/RSA_private.cpp
```diff
@@ -123,10 +123,10 @@ EVP_PKEY* RSA::d::GetKeyLowLevel() const { return m_pKey; }
 const EVP_PKEY* RSA::d::GetKey(const OTPasswordData* pPWData)
 {
     OT_ASSERT_MSG(
-        nullptr != backlink->m_p_ascKey,
+        !backlink->m_p_ascKey->Exists(),
         "RSA::GetKey: nullptr != m_p_ascKey\n");
```
FellowTraveler: You changed IF NOT NULL into IF DOESN'T EXIST

src/crypto/key/RSA_private.cpp
```diff
@@ -123,10 +123,10 @@ EVP_PKEY* RSA::d::GetKeyLowLevel() const { return m_pKey; }
...
 
-    if (nullptr == backlink->m_p_ascKey) {
+    if (backlink->m_p_ascKey->Exists()) {
         otErr << __FUNCTION__
```
FellowTraveler: You changed IF KEY IS NULL into IF KEY EXISTS

src/crypto/key/RSA_private.cpp
```diff
@@ -494,7 +494,7 @@ bool RSA::d::ArmorPublicKey(EVP_PKEY& theKey, Armored& ascKey)
 EVP_PKEY* RSA::d::InstantiatePublicKey(const OTPasswordData* pPWData)
 {
     OT_ASSERT(m_pKey == nullptr);
-    OT_ASSERT(backlink->m_p_ascKey != nullptr);
+    OT_ASSERT(backlink->m_p_ascKey->Exists());
     OT_ASSERT(backlink->IsPublic());
```
FellowTraveler: THIS ONE IS CORRECT!!!! GREAT JOB!!!!!!!!!!!!!!!!

src/crypto/key/RSA_private.cpp
```diff
@@ -561,7 +561,7 @@ EVP_PKEY* RSA::d::InstantiatePublicKey(const OTPasswordData* pPWData)
 EVP_PKEY* RSA::d::InstantiatePrivateKey(const OTPasswordData* pPWData)
 {
     OT_ASSERT(m_pKey == nullptr);
-    OT_ASSERT(backlink->m_p_ascKey != nullptr);
+    OT_ASSERT(backlink->m_p_ascKey->Exists());
     OT_ASSERT(backlink->IsPrivate());
```
FellowTraveler: **THIS ONE IS ALREADY CORRECT!!!! GREAT JOB ON THIS ONE!!!!!!!!**

PR #1273, opened 2018-11-06
Use LogOutput to replace otErr: Activity.cpp

Wallydog66:
Passes all unit tests.


tommyodom:
signed-off for grammar errors.

PR #1297, opened 2018-11-15
Use LogOutput to replace otErr: OTPayment.cpp

Wallydog66:
Passes all unit tests.


FellowTraveler:
This one looks good to me, go ahead and squash it with the rest.

Wallydog66:
Combined into one PR

PR #1299, opened 2018-11-17
Use LogOutput to replace otErr: MainFile.cpp

Wallydog66:
Passes all unit tests.


FellowTraveler:
This looks good to me. We can squash it in.

Wallydog66:
Combined into one PR

PR #1336, opened 2019-02-05
Fix build problems with OT_CASH

Wallydog66:
Signed-off-by: Kevin Waldren <wallydog.66@gmail.com>


src/api/client/Operation.cpp
```diff
@@ -2361,11 +2366,12 @@ bool Operation::SendCash(
 
     target_nym_id_ = recipientID;
     generic_id_ = workflowID;
+#if OT_CASH
```
tomlandrum: Delete lines 2369 and 2371.  They're redundant.

PR #1338, opened 2019-02-13
Use opentxs::identifier::Nym class for nym identifiers: V.2 with recommended changes

Wallydog66:
Use opentxs::identifier::Nym class for nym identifiers: V.2 with recommended changes


include/opentxs/core/OTTransactionType.hpp
```diff
@@ -51,8 +52,8 @@ class OTTransactionType : public Contract
     // Someday I'll add EntityID and RoleID here (in lieu of NymID,
     // in cases when the account is owned by an Entity and not a Nym.)
     //
-    inline const Identifier& GetNymID() const { return m_AcctNymID; }
-    inline void SetNymID(const Identifier& theID) { m_AcctNymID = theID; }
+    inline const identifier::Nym& GetNymID() const { return m_AcctNymID; }
+    inline void SetNymID(const identifier::Nym& theID) { m_AcctNymID = theID; }
```
FellowTraveler: I think this needs to be a Factory call here. like:

```{ m_AcctNymID = Factory(theID); }```

(Something like that.)

File: src/api/Wallet.cpp [hunk 2]
```diff
@@ -253,7 +254,7 @@ opentxs::Account* Wallet::account_factory(
         return nullptr;
     }
 
-    const auto signerNym = Nym(signerID);
+    const auto signerNym = Nym(identifier::Nym::Factory(signerID->str()));
```
FellowTraveler: Is signerID not already an identifier::Nym ?

src/api/Wallet.cpp
```diff
@@ -848,20 +850,22 @@ std::set<OTIdentifier> Wallet::IssuerList(const Identifier& nymID) const
 }
 
 std::shared_ptr<const api::client::Issuer> Wallet::Issuer(
-    const Identifier& nymID,
+    const identifier::Nym& nymID,
     const Identifier& issuerID) const
 {
-    auto& [lock, pIssuer] = issuer(nymID, issuerID, false);
+    auto& [lock, pIssuer] =
+        issuer(nymID, identifier::Nym::Factory(issuerID.str()), false);
```
FellowTraveler: Why not just pass issuerID (as the original version)?

Because if it's expecting an identifier::Nym, and if issuerID is already an identifier::Nym, that means it is already the correct type that's expected.

But if the type DOESN'T match, like if the issuer() call is expecting an Identifier instead of an identifier::Nym, then that would be the true problem (the mismatch). Just curious why the conversion to string and the new Factory call, unless issuerID is the wrong type? And if it's already the right type, then probably the issuer() call needs to be fixed to expect that type, rather than fixing the argument passed to it.

src/api/Wallet.cpp
```diff
@@ -848,20 +850,22 @@ std::set<OTIdentifier> Wallet::IssuerList(const Identifier& nymID) const
...
 }
 
 Editor<api::client::Issuer> Wallet::mutable_Issuer(
-    const Identifier& nymID,
+    const identifier::Nym& nymID,
     const Identifier& issuerID) const
 {
-    auto& [lock, pIssuer] = issuer(nymID, issuerID, true);
+    auto& [lock, pIssuer] =
+        issuer(nymID, identifier::Nym::Factory(issuerID.str()), true);
```
FellowTraveler: Here I can CLEARLY see the exact problem I pointed out above...

Notice in the function arguments, the type for issuerID is the OLD type:
```const Identifier& issuerID)```

If the issuerID argument is in fact supposed to be a NYM ID, then it should be changed to:
```const identifier::Nym& issuerID)```

...And once you do that, then you would call the issuer() call like this:

```issuer(nymID, issuerID, true);```

(Which is exactly how the original code looked).

If that gives you an error, it would be because the issuer() call itself is still expecting an Identifier for its second argument, instead of an identifier::Nym as it should be. Once you fix the issuer() call to take the correct type, then the above code that calls it will work without the string conversion and without the Factory.

File: src/api/Wallet.cpp [hunk 5]
```diff
@@ -1818,7 +1824,7 @@ void Wallet::save(
 
     OT_ASSERT(false == signerID->empty())
 
-    const auto signerNym = Nym(signerID);
+    const auto signerNym = Nym(identifier::Nym::Factory(signerID->str()));
```
FellowTraveler: If signerID is not a NymID, then you need to change it to a NymID, wherever it is declared.

Once you do that, you will no longer need the Factory call here. Instead you can call it as it was called originally:
```const auto signerNym = Nym(signerID);```

File: src/api/Wallet.cpp [hunk 6]
```diff
@@ -2119,7 +2126,7 @@ ConstServerContract Wallet::Server(
 {
     std::string server;
 
-    auto nym = Nym(Identifier::Factory(nymid));
+    auto nym = Nym(identifier::Nym::Factory(nymid));
```
FellowTraveler: What is ```nymid``` here? If it's an identifier::Nym already, then no Factory call is necessary.

Whereas if it's an Identifier, then the solution is to fix it to be an identifier::Nym.

However, **if nymid is a string,** then your code is probably already correct.

src/api/Wallet.cpp
```diff
@@ -2157,14 +2165,14 @@ OTIdentifier Wallet::server_to_nym(OTIdentifier& input) const
 
         for (const auto& item : list) {
             const auto& serverID = item.first;
-            auto server = Server(Identifier::Factory(serverID));
+            auto server = Server(identifier::Nym::Factory(serverID));
```
FellowTraveler: This is a bug. This needs to be changed back to the original code.

I guarantee you that Server is expecting a server ID (Identifier) and **not** a Nym ID (identifier::Nym). So you should continue passing it the right type, and not change it to pass the wrong type as you have done here.

src/api/Wallet.cpp
```diff
@@ -2157,14 +2165,14 @@ OTIdentifier Wallet::server_to_nym(OTIdentifier& input) const
...
 
             OT_ASSERT(server);
 
-            if (server->Nym()->ID() == input) {
+            if (server->Nym()->ID().operator==(input)) {
```
FellowTraveler: Presumably ```server->Nym()->ID()``` is of type identifier::Nym.

Yet you are COMPARING it to input, which is of type OTIdentifier.

Therefore, unless identifier::Nym has a special ```operator==``` function that takes an Identifier&, you otherwise appear to be comparing apples and oranges. (Although such a function may very well exist, allowing the two types to compare to each other).

A better move here would be to just make sure that both parameters ARE THE SAME TYPE, and then compare them. That's the only thing that will work, unless as I said there is some special function on one of those classes which enables it to compare itself to other types.

src/api/Wallet.cpp
```diff
@@ -2157,14 +2165,14 @@ OTIdentifier Wallet::server_to_nym(OTIdentifier& input) const
...
                 matches++;
                 // set input to the notary ID
-                input = Identifier::Factory(server->ID());
+                input->str() = server->ID()->str();
```
FellowTraveler: This is a bad bug here.

I think you should stick with the original code, which looks correct:
```input = Identifier::Factory(server->ID());```

If that's not working for you, then try this:
```input = Identifier::Factory(server->ID()->str());```

But you definitely should NOT be assigning any values into input->str().

To be clear, input->str() is for RETRIEVING a string copy of the input variable. It's NOT a function for assigning a value. You should never be writing into the str() that's retrieved. That string is for reading purposes only, not writing.

File: src/api/Wallet.cpp [hunk 10]
```diff
@@ -2173,15 +2181,15 @@ OTIdentifier Wallet::server_to_nym(OTIdentifier& input) const
         auto contract = Server(input);
 
         if (contract) {
-            output = Identifier::Factory(contract->Contract().nymid());
+            output = identifier::Nym::Factory(contract->Contract().nymid());
```
FellowTraveler: I just wanted to point out here that you are copying the wrong type...

Notice the **type** for output: ```auto output = Identifier::Factory();```

(So it's an OTIdentifier).

But notice what you are COPYING into the output variable...an identifier::Nym, which is a different type than output:
```output = identifier::Nym::Factory(contract->Contract().nymid());```

You CANNOT copy an identifier::Nym into an OTIdentifier. They are two different types. You need to either convert the identifier::Nym value to an Identifier object before you assign it to output (which is also an Identifier object), OR you need to change the output variable to the new type (identifier::Nym) so that when you copy the identifier::Nym into it, it works since they are then the same type.

You can't be copying the wrong type into objects, especially with these Factories where maybe it allows a build but then causes a memory problem later on at runtime.

These kinds of bugs are the ones that will cause your tests to fail...

File: src/api/Wallet.cpp [hunk 11]
```diff
@@ -2173,15 +2181,15 @@ OTIdentifier Wallet::server_to_nym(OTIdentifier& input) const
...
         } else {
             LogDetail(OT_METHOD)(__FUNCTION__)(": Non-existent server: ")(input)
                 .Flush();
         }
     }
 
-    return output;
-}
+    return identifier::Nym::Factory(output->str());
```
FellowTraveler: Here you are returning an identifier::Nym.

But notice the actual RETURN TYPE of this function:
```OTIdentifier Wallet::server_to_nym(OTIdentifier& input) const```

So here you are returning an identifier::Nym to the caller, who receives it into an OTIdentifier object.

THE TYPES MUST MATCH. You cannot return a different type than the one defined in the function definition.

If you DO want to return an identifier::Nym here (which is probably the right move) then you need to ALSO change the return type for the function itself, like this:

```OTNymID Wallet::server_to_nym(OTIdentifier& input) const```

...and if you do that, make sure you change the appropriate headers as well, not just the CPP file.

File: src/api/Wallet.hpp [hunk 12]
```diff
@@ -217,14 +220,14 @@ class Wallet : virtual public opentxs::api::Wallet, public Lockable
...
+        const identifier::Nym& localNymID,
+        const identifier::Nym& remoteNymID) const;
     proto::ContactItemType extract_unit(const Identifier& contractID) const;
     proto::ContactItemType extract_unit(
         const opentxs::UnitDefinition& contract) const;
-    void save(opentxs::internal::Context* context) const;
-    OTIdentifier server_to_nym(OTIdentifier& serverID) const;
 
+    void save(opentxs::internal::Context* context) const;
+    OTNymID server_to_nym(OTIdentifier& serverID) const;
```
FellowTraveler: Looks here like you DID change the return value of the function in the header (which is good) but if it's the same function I commented above, you didn't make the corresponding change in the CPP file.

src/api/client/Activity.cpp
```diff
@@ -419,7 +420,7 @@ std::string Activity::Mail(
         std::thread preload(
             &Activity::preload,
             this,
-            Identifier::Factory(nym),
+            identifier::Nym::Factory(nym.str()),
```
FellowTraveler: What IS the "nym" object here? What TYPE is it?

I thought it was a string, but in your new code, I see you converting it to a string in order to pass it to the factory.

So is it an actual Nym object? In which case you probably want to pass something like ```nym->ID()->str()``` to the Factory call. Otherwise you might be retrieving a string copy of the entire Nym, versus a string copy of his ID. And that would result in your factory call creating an invalid ID object which it then passes to the preload function.

I'd have to see the actual type (of the nym object) to know more.

src/api/client/Contacts.cpp
```diff
@@ -122,12 +123,12 @@ void Contacts::check_identifiers(
 
     if (false == inputNymID.empty()) {
         haveNymID = true;
-        outputNymID = Identifier::Factory(inputNymID);
+        outputNymID = inputNymID;
```
FellowTraveler: I would think you need to do it like this:
```outputNymID = identifier::Nym::Factory(inputNymID);```

(Because outputNymID is an OTNymID...)

File: src/api/client/Issuer.cpp [hunk 15]
```diff
@@ -48,29 +49,29 @@ namespace opentxs::api::client::implementation
...
+    const identifier::Nym& nymID,
     const Identifier& issuerID)
     : wallet_(wallet)
     , version_(CURRENT_VERSION)
     , pairing_code_("")
     , paired_(Flag::Factory(false))
-    , nym_id_(Identifier::Factory(nymID))
-    , issuer_id_(Identifier::Factory(issuerID))
+    , nym_id_(nymID)
+    , issuer_id_(issuerID)
```
FellowTraveler: The old code was using a Factory, so I'd expect the new code to use a factory as well.

You probably should have kept this old line:
```, issuer_id_(Identifier::Factory(issuerID))```

And probably for the NymID should be something like this:
```, nym_id_(identifier::Nym::Factory(nymID))``` possibly with a ->str() call in there.

File: src/api/client/Issuer.cpp [hunk 16]
```diff
@@ -48,29 +49,29 @@ namespace opentxs::api::client::implementation
...
+    const identifier::Nym& nymID,
     const proto::Issuer& serialized)
     : wallet_(wallet)
     , version_(serialized.version())
     , pairing_code_(serialized.pairingcode())
     , paired_(Flag::Factory(serialized.paired()))
-    , nym_id_(Identifier::Factory(nymID))
-    , issuer_id_(Identifier::Factory(serialized.id()))
+    , nym_id_(nymID)
+    , issuer_id_(identifier::Nym::Factory(serialized.id()))
```
FellowTraveler: Here you changed the issuer_id line from this:
```, issuer_id_(Identifier::Factory(serialized.id()))```

To this:
```, issuer_id_(identifier::Nym::Factory(serialized.id()))```

So my question for you is, IS the ```issuer_id_``` variable an OTNymID? Is that the type? If so, fair enough. But otherwise, if it's an OTIdentifier, then the original code was correct.

As for nym_id, the original code was:
```, nym_id_(Identifier::Factory(nymID))```

I would expect the new code should be:
```, nym_id_(identifier::Nym::Factory(nymID))```

or:
```, nym_id_(identifier::Nym::Factory(nymID->str()))```

I would have to check the header to see the **actual** types for ```nym_id_``` and ```issuer_id_``` (and THOSE are what dictate which factory to use, etc...)

File: src/api/client/OTX.cpp [hunk 17]
```diff
@@ -1416,19 +1416,21 @@ bool OTX::find_server(
...
 
-OTX::BackgroundTask OTX::FindNym(const Identifier& nymID) const
+OTX::BackgroundTask OTX::FindNym(const identifier::Nym& nymID) const
 {
     CHECK_NYM(nymID)
 
     const auto taskID{++next_task_id_};
     // TODO convert nym id type
-    const auto id = identifier::Nym::Factory(nymID.str());
+    const auto id = nymID.str();
```
FellowTraveler: ABOVE, the old code shows that the **id** variable is an OTNymID.
Whereas the new code shows that the id variable is a STRING.

So my question for you is, which one is it SUPPOSED to be? A NymID, or a String containing one?

Probably could have done this:
```const auto id = identifier::Nym::Factory(nymID.str());```

Or this:
```const auto id = identifier::Nym::Factory(nymID.str().c_str());```

And then you could have probably just kept the original line below like this:
```return start_task(taskID, missing_nyms_.Push(taskID, id));`

File: src/api/client/OTX.cpp [hunk 18]
```diff
@@ -2511,7 +2515,7 @@ OTX::BackgroundTask OTX::SendExternalTransfer(
         return error_task();
     }
 
-    if (sourceAccount.get().GetNymID() != localNymID) {
+    if (sourceAccount.get().GetNymID().operator!=(localNymID)) {
```
FellowTraveler: What TYPE is returned by the GetNymID() function?

What TYPE is the localNymID variable?

Are they the SAME TYPE?

If so, then it's fine comparing them.

But otherwise, you cannot compare them without first converting one of the types (so they match) or alternately by making sure that the types have special operator!= and operator== functions that enable them to compare themselves to the different type.

src/api/client/OTX.cpp
```diff
@@ -2531,11 +2535,15 @@ OTX::BackgroundTask OTX::SendExternalTransfer(
     return start_task(
         taskID,
         queue.send_transfer_.Push(
-            taskID, {sourceAccountID, targetAccountID, value, memo}));
+            taskID,
+            {identifier::Nym::Factory(sourceAccountID.str()),
```
FellowTraveler: This looks to be an account ID, NOT a Nym ID.

So I don't think you should be converting it to a Nym ID...

Also, this is send_transfer, which is known for sending FROM an account ID, TO an account ID _(not from/to a NymID)._

src/api/client/OTX.cpp
```diff
@@ -2553,7 +2561,7 @@ OTX::BackgroundTask OTX::SendTransfer(
         return error_task();
     }
 
-    if (sourceAccount.get().GetNymID() != localNymID) {
+    if (sourceAccount.get().GetNymID().operator!=(localNymID)) {
```
FellowTraveler: See above notes about making sure these are the same type, and otherwise making sure, if they are different types, that they have the appropriate function to compare themselves to each other.

src/api/client/OTX.cpp
```diff
@@ -2573,10 +2581,15 @@ OTX::BackgroundTask OTX::SendTransfer(
     return start_task(
         taskID,
         queue.send_transfer_.Push(
-            taskID, {sourceAccountID, targetAccountID, value, memo}));
+            taskID,
+            {identifier::Nym::Factory(sourceAccountID.str()),
```
FellowTraveler: Transfers use account IDs, not NymIDs...

src/api/client/OTX.hpp
```diff
@@ -300,7 +301,7 @@ class OTX : virtual public client::OTX, Lockable
     /** SendTransferTask: localNymID, serverID, sourceAccountID, targetAccountID
      */
     using SendTransferTask =
-        std::tuple<OTIdentifier, OTIdentifier, uint64_t, std::string>;
```
FellowTraveler: Something is wrong here. The doxygen comment (documentation) says:
```SendTransferTask: localNymID, serverID, sourceAccountID, targetAccountID```

That should be:
```std::tuple<OTNymID, OTIdentifier, std::string, std::string>;```

But instead I see this:
```std::tuple<OTNymID, OTIdentifier, uint64_t, std::string>;```

Notice the first account ID is now a uint64_t (a form of integer) when we all know that both account IDs are expected to be strings. At the very least, both account IDs should be the same type. It's impossible that one of them is an integer while the other is a string.

Likely the integer is meant to be an amount. Something like this:
```SendTransferTask: sourceAccountID, targetAccountID, amount, serverID,```

Honestly I don't know, I'm guessing. I would look at where SendTransferTask is ACTUALLY defined, and make sure it looks correct there, and THEN make sure that the code here matches those types. Because right now something is funny and the doxygen comment is obviously outdated.

File: src/api/client/Operation.cpp [hunk 23]
```diff
@@ -687,11 +689,11 @@ std::shared_ptr<Message> Operation::construct_deposit_cheque()
     if (cheque.HasRemitter()) {
         cancellingCheque =
             ((cheque.GetRemitterAcctID() == account_id_) &&
-             (cheque.GetRemitterNymID() == nymID));
+             (cheque.GetRemitterNymID().operator==(nymID)));
```
FellowTraveler: Just make sure that GetRemitterNymID() returns the same type as nymID's type. The types need to be the same. If they are not, convert one of them so that they match.

Alternately, make sure if they ARE two different types, that they know how to compare to each other, so that the comparison operator works properly even if they are different types. Verify that operator function exists and takes the right types.

src/api/client/Operation.cpp
```diff
@@ -687,11 +689,11 @@ std::shared_ptr<Message> Operation::construct_deposit_cheque()
...
     } else {
         cancellingCheque =
             ((cheque.GetSenderAcctID() == account_id_) &&
-             (cheque.GetSenderNymID() == nymID));
+             (cheque.GetSenderNymID().operator==(nymID)));
```
FellowTraveler: same thing here.

src/api/client/ServerAction.cpp
```diff
@@ -277,12 +278,12 @@ ServerAction::Action ServerAction::KillPaymentPlan(
         api_,
         localNymID,
         serverID,
-        accountID,
+        identifier::Nym::Factory(accountID.str()),
```
FellowTraveler: Why are you converting the accountID variable into an OTNymID type?
Look at the function you are calling, wherever header it's defined -- are you SUPPOSED to be passing a NymID in the place where it expects an account ID? Is that how it's defined?

Otherwise the old code is probably more correct...

src/api/client/ServerAction.cpp
```diff
@@ -295,14 +296,14 @@ ServerAction::Action ServerAction::PayDividend(
...
```
FellowTraveler: See above comment.

src/api/client/ServerAction.cpp
```diff
@@ -330,11 +331,11 @@ ServerAction::Action ServerAction::UnregisterAccount(
         api_,
         localNymID,
         serverID,
-        accountID));
+        identifier::Nym::Factory(accountID.str())));
```
FellowTraveler: See above comment...

src/api/client/ServerAction.cpp
```diff
@@ -359,7 +360,7 @@ ServerAction::Action ServerAction::WithdrawVoucher(
         api_,
         localNymID,
         serverID,
-        accountID,
+        identifier::Nym::Factory(accountID.str()),
```
FellowTraveler: See above comment..

src/api/client/Workflow.cpp
```diff
@@ -509,7 +510,7 @@ bool Workflow::AcceptTransfer(
     const auto recipientNymID = pending.GetNymID().str();
     const auto& accountID = pending.GetPurportedAccountID();
 
-    if (pending.GetNymID() != nymID) {
+    if (pending.GetNymID().operator==(nymID)) {
```
FellowTraveler: This looks like bad logic.

The old code was verifying that the IDs did NOT match.

Whereas your code here verifies that they DO match.

Which is the opposite of what the old code did.

Separately, you also want to make sure you are comparing objects of the same type, or that alternately they know how to compare themselves to each other, even when they are different types. (Custom operator== and operator!= functions make this possible, and many classes have them where needed).

src/api/client/Workflow.cpp
```diff
@@ -1541,7 +1542,7 @@ OTIdentifier Workflow::ConveyTransfer(
     contact_.NymToContact(transfer->GetNymID());
     const auto recipientNymID = pending.GetNymID().str();
 
-    if (pending.GetNymID() != nymID) {
+    if (pending.GetNymID().operator==(nymID)) {
```
FellowTraveler: Notice again the logic reversal here. It cannot possible be right! This code logs an "invalid recipient" error in the case where they MATCH. Versus when they fail to match.

src/api/storage/Storage.cpp
```diff
@@ -381,13 +381,15 @@ std::set<OTIdentifier> Storage::AccountsByContract(
 std::set<OTIdentifier> Storage::AccountsByIssuer(
     const Identifier& issuerNym) const
```
FellowTraveler: Looks to me like the issuerNym parameter should be changed to: ```const identifier::Nym& issuerNym```

...and once you do that, you no longer have to put the Factory call below...

src/api/storage/Storage.cpp
```diff
@@ -381,13 +381,15 @@ std::set<OTIdentifier> Storage::AccountsByContract(
...
 {
-    return Root().Tree().Accounts().AccountsByIssuer(issuerNym);
+    return Root().Tree().Accounts().AccountsByIssuer(
+        identifier::Nym::Factory(issuerNym.str()));
 }
 
 std::set<OTIdentifier> Storage::AccountsByOwner(
     const Identifier& ownerNym) const
```
FellowTraveler: Same thing here as above.

src/api/storage/Storage.cpp
```diff
@@ -1908,9 +1910,9 @@ bool Storage::Store(
             accountID,
             data,
             alias,
-            ownerNym,
-            signerNym,
-            issuerNym,
+            identifier::Nym::Factory(ownerNym.str()),
+            identifier::Nym::Factory(signerNym.str()),
+            identifier::Nym::Factory(issuerNym.str()),
             server,
```
FellowTraveler: What function is being called here?

Looks like it now expects an OTNymID, so you are using Factory calls to give it what it wants. (Fair enough).

But then why not just change the ownerNym, signerNym, and issuerNym variables to be the correct type in the first place, so you don't have to convert them to OTNymID later when passing those variables to function calls?

src/client/OTAPI_Exec.cpp
```diff
@@ -3999,18 +4002,17 @@ std::string OTAPI_Exec::WriteCheque(
...
         theRecipientNymID->SetString(String::Factory(RECIPIENT_NYM_ID));
     auto strMemo = String::Factory();
 
     if (!CHEQUE_MEMO.empty()) strMemo->Set(String::Factory(CHEQUE_MEMO));
 
-    OTIdentifier idForRecipient(
-        bHasRecipient ? Identifier::Factory(theRecipientNymID)
-                      : Identifier::Factory());
+    OTNymID idForRecipient(
+        bHasRecipient ? theRecipientNymID : identifier::Nym::Factory());
```
FellowTraveler: ```theRecipientNymID``` should probably be: ```identifier::Nym::Factory(theRecipientNymID)```

src/client/OT_API.cpp
```diff
@@ -5704,9 +5708,9 @@ bool OT_API::RecordPayment(
                             // case, he's the "recipient." (So we need to figure
                             // out which, and set the account accordingly.)
                             //
-                            if (NYM_ID == pPlan->GetRecipientNymID())
+                            if (NYM_ID.operator==(pPlan->GetRecipientNymID()))
                                 theSenderAcctID = pPlan->GetRecipientAcctID();
-                            else if (NYM_ID == pPlan->GetSenderNymID())
+                            else if (NYM_ID.operator==(pPlan->GetSenderNymID()))
                                 theSenderAcctID = pPlan->GetSenderAcctID();
```
FellowTraveler: Presumably whenever you do the comparison operator (operator==) you are comparing objects of the same type.

If they are not the same type, you must do one of two things:

1. You can convert one of them so they ARE the same type, then compare them. Or,
2. You can make sure the objects have special operator== functions that enable them to compare themselves to each other. There must be special functions in place for each different type it needs to be comparable to.

Just making sure you verified this when you made this change.

src/client/OT_API.cpp
```diff
@@ -7182,7 +7186,7 @@ CommandResult OT_API::payDividend(
         SHARES_ISSUER_accountID,
         nymID,
         DIVIDEND_MEMO,
-        Identifier::Factory(nymID));
+        nymID);
```
FellowTraveler: Why didn't you change this to identifier::Nym::Factory(nymID)); ?

Just curious what I'm missing, since the original call used the old Factory, so I'd normally expect the fixed version to use the new Factory. But instead you aren't using the Factory at all. Just curious why not.

src/client/SwigWrap.cpp
```diff
@@ -2984,18 +2987,20 @@ std::string SwigWrap::Add_Contact(
 
     if (noLabel && noNym && noPaymentCode) { return {}; }
 
-    auto nym = Identifier::Factory(nymID);
+    auto nym = nymID;
```
FellowTraveler: Shouldn't this be:
```auto nym = identifier::Nym::Factory(nymID);```

File: src/client/SwigWrap.cpp [hunk 38]
```diff
@@ -2984,18 +2987,20 @@ std::string SwigWrap::Add_Contact(
...
-    if (nym->empty() && code->VerifyInternally()) { nym = code->ID(); }
+    if (nym.empty() && code->VerifyInternally()) {
+        nym.operator=(code->ID()->str());
+    }
 #endif
 
     auto output = client_->Contacts().NewContact(
         label,
-        nym
+        identifier::Nym::Factory(nym)
```
FellowTraveler: If the old code didn't use the Factory, then why does the new code?

Probably just need to make sure that ```nym``` is the correct type.

src/contact/Contact.cpp
```diff
@@ -38,9 +39,9 @@ Contact::Contact(const api::Wallet& wallet, const proto::Contact& serialized)
     , version_(check_version(serialized.version(), OT_CONTACT_VERSION))
     , label_(serialized.label())
     , lock_()
-    , id_(Identifier::Factory(Identifier::Factory(serialized.id())))
+    , id_(Identifier::Factory(serialized.id()))
```
FellowTraveler: If the old code used the old Factory, why doesn't the new code use the new Factory? Just curious...

src/core/Cheque.cpp
```diff
@@ -278,7 +279,7 @@ bool Cheque::IssueCheque(
         m_RECIPIENT_NYM_ID->Release();
     } else {
         m_bHasRecipient = true;
-        m_RECIPIENT_NYM_ID = Identifier::Factory(pRECIPIENT_NYM_ID);
+        m_RECIPIENT_NYM_ID = pRECIPIENT_NYM_ID;
```
FellowTraveler: Surprised this new code doesn't use identifier::Nym::Factory

src/core/Ledger.cpp
```diff
@@ -1004,15 +1005,15 @@ bool Ledger::GenerateLedger(
...
+    auto nymID = identifier::Nym::Factory();
 
     if ((ledgerType::inbox == theType) || (ledgerType::outbox == theType)) {
         // Have to look up the NymID here. No way around it. We need that ID.
         // Plus it helps verify things.
         auto account = api_.Wallet().Account(theAcctID);
 
         if (account) {
-            nymID = Identifier::Factory(account.get().GetNymID());
+            nymID = (account.get().GetNymID());
```
FellowTraveler: Does GetNymID() return a Factory call?

If not, I don't understand why you didn't just keep the Factory call here, except as the new version, like this:
```nymID = identifier::Nym::Factory(account.get().GetNymID());```

File: src/core/Ledger.cpp [hunk 42]
```diff
@@ -1025,26 +1026,26 @@ bool Ledger::GenerateLedger(
...
```
FellowTraveler: See above comment.

src/core/Nym.cpp
```diff
@@ -81,7 +82,7 @@ Nym::Nym(
     , mode_(mode)
     , m_strVersion(String::Factory(NYMFILE_VERSION))
     , m_strDescription(String::Factory())
-    , m_nymID(Identifier::Factory(nymID))
+    , m_nymID(nymID)
```
FellowTraveler: Probably should have put the new Factory call here, instead of removing Factory entirely.
Seems like a bad initialization to me...

src/core/NymIDSource.cpp
```diff
@@ -146,7 +147,7 @@ OTIdentifier NymIDSource::NymID() const
             break;
 #if OT_CRYPTO_SUPPORTED_SOURCE_BIP47
         case proto::SOURCETYPE_BIP47:
-            nymID = payment_code_->ID();
+            nymID->operator==(payment_code_->ID());
```
FellowTraveler: This is a bad bug. You changed an _assignment_ to use a _comparison_ instead. (And the result of the comparison is then discarded...)

Probably you meant to do this:
```nymID->operator=(payment_code_->ID());```

Or:
```nymID = identifier::Nym::Factory(payment_code_->ID());```

Or:
```nymID = identifier::Nym::Factory(payment_code_->ID()->str());```

File: src/core/OTTrackable.cpp [hunk 45]
```diff
@@ -91,9 +92,9 @@ void OTTrackable::SetSenderAcctID(const Identifier& ACCT_ID)
     m_SENDER_ACCT_ID = Identifier::Factory(ACCT_ID);
 }
 
-void OTTrackable::SetSenderNymID(const Identifier& NYM_ID)
+void OTTrackable::SetSenderNymID(const identifier::Nym& NYM_ID)
 {
-    m_SENDER_NYM_ID = Identifier::Factory(NYM_ID);
+    m_SENDER_NYM_ID = (NYM_ID);
```
FellowTraveler: Why did you not update to the new Factory call? Like this:

```m_SENDER_NYM_ID = identifier::Nym::Factory(NYM_ID);```

File: src/core/OTTransactionType.cpp [hunk 46]
```diff
@@ -52,15 +53,15 @@ OTTransactionType::OTTransactionType(const api::Core& core)
...
+    const identifier::Nym& theNymID,
     const Identifier& theAccountID,
     const Identifier& theNotaryID,
     originType theOriginType)
     : Contract(core, theAccountID)
     , m_AcctID(Identifier::Factory())
     , m_NotaryID(Identifier::Factory(theNotaryID))
     , m_AcctNotaryID(Identifier::Factory())
-    , m_AcctNymID(Identifier::Factory(theNymID))
+    , m_AcctNymID(theNymID)
```
FellowTraveler: This also should be using the new Factory call.

src/core/OTTransactionType.cpp
```diff
@@ -84,7 +85,7 @@ OTTransactionType::OTTransactionType(
...
```
FellowTraveler: Same thing here...

src/core/cron/OTCronItem.cpp
```diff
@@ -895,8 +896,7 @@ void OTCronItem::HookRemovalFromCron(
             // but for whatever reason, I'm checking the nymID on the original
             // version. Sue me.
             //
-            const OTIdentifier NYM_ID =
-                Identifier::Factory(pOrigCronItem->GetSenderNymID());
+            const identifier::Nym& NYM_ID = (pOrigCronItem->GetSenderNymID());
```
FellowTraveler: This is really screwed up. The original line:
```const OTIdentifier NYM_ID = Identifier::Factory(pOrigCronItem->GetSenderNymID());```

Should have been changed to something like this:
```const OTNymID NYM_ID = identifier::Nym::Factory(pOrigCronItem->GetSenderNymID());```

Instead, you changed it to a reference...

File: src/core/cron/OTCronItem.cpp [hunk 49]
```diff
@@ -1008,7 +1008,7 @@ void OTCronItem::onFinalReceipt(
         context.It().VerifyIssuedNumber(lClosingNumber)) {
         // SENDER only. (CronItem has no recipient. That's in the subclass.)
         if (!DropFinalReceiptToInbox(
-                GetSenderNymID(),
+                identifier::Nym::Factory(),
```
FellowTraveler: Here you changed it from passing the SenderNymID, to instead passing a blank ID...

That can't be right...

src/core/identifier/CMakeLists.txt
```diff
@@ -9,9 +9,16 @@
...
      "${CMAKE_CURRENT_SOURCE_DIR}/../../../include/opentxs/core/identifier/Nym.hpp"
      "${CMAKE_CURRENT_SOURCE_DIR}/../../../include/opentxs/core/identifier/Server.hpp"
      "${CMAKE_CURRENT_SOURCE_DIR}/../../../include/opentxs/core/identifier/UnitDefinition.hpp"
+=======
+  "${CMAKE_CURRENT_SOURCE_DIR}/../../../include/opentxs/core/identifier/Nym.hpp"
+  "${CMAKE_CURRENT_SOURCE_DIR}/../../../include/opentxs/core/identifier/Server.hpp"
+  "${CMAKE_CURRENT_SOURCE_DIR}/../../../include/opentxs/core/identifier/UnitDefinition.hpp"
+
+>>>>>>> e8812e4f0... Use opentxs::identifier::Nym class for nym identifiers
 )
```
FellowTraveler: This is a merge artifact. Definitely needs to be fixed before it will work.

src/ext/OTPayment.cpp
```diff
@@ -140,9 +141,9 @@ OTPayment::paymentType OTPayment::GetTypeFromString(const String& strType)
     return OTPayment::ERROR_STATE;
 }
 
-bool OTPayment::SetTempRecipientNymID(const Identifier& id)
+bool OTPayment::SetTempRecipientNymID(const identifier::Nym& id)
 {
-    m_RecipientNymID = Identifier::Factory(id);
+    m_RecipientNymID = id;
```
FellowTraveler: Shouldn't this be identifier::Nym::Factory() ?

src/server/PayDividendVisitor.cpp
```diff
@@ -34,14 +34,14 @@ namespace opentxs
...
-    const Identifier& theNymID,
+    const identifier::Nym& theNymID,
     const Identifier& thePayoutUnitTypeId,
     const Identifier& theVoucherAcctID,
     const String& strMemo,
     std::int64_t lPayoutPerShare)
     : AccountVisitor(server.API().Wallet(), Identifier::Factory(theNotaryID))
     , server_(server)
-    , nymId_(Identifier::Factory(theNymID))
+    , nymId_(theNymID)
```
FellowTraveler: Shouldn't this have the new Factory call?

src/ui/ActivitySummaryItem.cpp
```diff
@@ -63,14 +64,14 @@ ActivitySummaryItem::ActivitySummaryItem(
...
-    const Identifier& nymID,
+    const identifier::Nym& nymID,
     const ActivitySummaryRowID& rowID,
     const ActivitySummarySortKey& sortKey,
     const CustomData& custom,
     const Flag& running)
     : ActivitySummaryItemRow(parent, api, publisher, rowID, true)
     , running_(running)
-    , nym_id_(Identifier::Factory(nymID))
+    , nym_id_(nymID)
```
FellowTraveler: Shouldn't this be the new Factory call?

src/ui/Contact.cpp
```diff
@@ -64,7 +66,7 @@ const std::map<proto::ContactSectionName, int> Contact::sort_keys_{
 Contact::Contact(
     const api::client::Manager& api,
     const network::zeromq::PublishSocket& publisher,
-    const Identifier& contactID)
+    const identifier::Nym& contactID)
```
FellowTraveler: Are we supposed to be using NymIDs for ContactIDs??

PR #1343, opened 2019-02-20
Updates for building on macos.

tomlandrum:


src/core/Message.cpp
```diff
@@ -3669,12 +3669,6 @@ class StrategyGetMintResponse : public OTMessageStrategy
         const auto found = String::Factory(xml->getAttributeValue("found"));
         m.m_bBool = found->Compare("true");
 
-        try {
-            m.enum_ = std::stoi(xml->getAttributeValue("type"));
-        } catch (...) {
-            m.enum_ = 0;
-        }
-
```
justusranvier: Why was this removed?

src/network/zeromq/socket/Socket.cpp
```diff
@@ -50,6 +50,11 @@ Socket::Socket(
     , running_(Flag::Factory(true))
     , type_(type)
 {
+	if (nullptr == socket_) {
+        LogOutput(OT_METHOD)(__FUNCTION__)(": ")(zmq_strerror(zmq_errno()))
+            .Flush();
+	}
+
```
justusranvier: It's not safe to use Log class here

tests/OTTestEnvironment.cpp
```diff
@@ -6,10 +6,17 @@
 #include "OTTestEnvironment.hpp"
 
 #include "opentxs/opentxs.hpp"
+#include "core/String.hpp"
```
justusranvier: should be "opentxs/core/String.hpp"

src/core/Message.cpp
tomlandrum: It was causing an unhandled exception on macos.  Since that attribute isn't set in writeXml, I thought it would be okay to just remove it from processXml.

src/network/zeromq/socket/Socket.cpp
tomlandrum: I knew that.

tests/OTTestEnvironment.cpp
justusranvier: Due to how the include paths are set in cmake, #include "core/String.hpp" will match the private header in src/core rather than the public header in include/opentxs/core

src/core/Message.cpp
tomlandrum: It was a segfault, not an unhandled exception.

PR #1478, opened 2020-05-17
Added Verify and Subscribe states to Peer machine.

FellowTraveler:
Signed-off-by: FellowTraveler <F3llowTraveler@gmail.com>

### DIFF
--- a/src/blockchain/client/HeaderOracle.cpp
+++ b/src/blockchain/client/HeaderOracle.cpp
@@ -82,21 +82,24 @@ auto HeaderOracle::GenesisBlockHash(const blockchain::Type type)
 
 namespace opentxs::blockchain::client::implementation
 {
-const std::map<blockchain::Type, std::pair<block::Height, std::string>>
-    HeaderOracle::checkpoints_{
-        {blockchain::Type::Bitcoin,
-         {630000,
-          "6de9d737a62ea1c197000edb02c252089969dfd8ea4b02000000000000000000"}},
-        {blockchain::Type::Bitcoin_testnet3,
-         {1740000,
-          "f758e6307382affd044c64e2bead2efdd2d9222bce9d232ae3fc000000000000"}},
-        {blockchain::Type::BitcoinCash,
-         {635259,
-          "f73075b2c598f49b3a19558c070b52d5a5d6c21fefdf33000000000000000000"}},
-        {blockchain::Type::BitcoinCash_testnet3,
-         {1378461,
-          "d715e9fab7bbdf301081eeadbe6e931db282cf6b92b1365f9b50f59900000000"}},
-    };
+const HeaderOracle::CheckpointsType HeaderOracle::checkpoints_{
+    {blockchain::Type::Bitcoin,
+     {630000,
+      "6de9d737a62ea1c197000edb02c252089969dfd8ea4b02000000000000000000",
+      "1e2b96b120a73bacb0667279bd231bdb95b08be16b650d000000000000000000"}},
+    {blockchain::Type::Bitcoin_testnet3,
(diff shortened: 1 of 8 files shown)


src/blockchain/client/HeaderOracle.cpp
```diff
@@ -506,33 +507,39 @@ auto HeaderOracle::GetCheckpoint() const noexcept -> block::Position
...
     }
 
     LogNormal(blockchain::internal::DisplayString(chain_))(
-        ": Updating checkpoint to hash ")(expected.second->asHex())(
-        " at height ")(expected.first)
+        ": Updating checkpoint to hash ")(defaultBlockhash)(" at height ")(
+        defaultHeight)
         .Flush();
-    const auto added = AddCheckpoint(expected.first, expected.second);
+    auto dataDefaultBlockhash = api_.Factory().Identifier();
```
justusranvier, line 539: This works but it should be a Data instead of an Identifier.

src/blockchain/client/HeaderOracle.hpp
```diff
@@ -51,12 +52,16 @@ namespace opentxs::blockchain::client::implementation
 class HeaderOracle final : virtual public internal::HeaderOracle
 {
 public:
+    using CheckpointData = std::tuple<block::Height, std::string, std::string>;
```
justusranvier, line 55: Instead of strings, the public functions should return block::pHash

src/blockchain/p2p/bitcoin/Peer.cpp
```diff
@@ -693,29 +693,63 @@ auto Peer::process_headers(
     if (false == bool(pMessage)) {
         LogOutput(OT_METHOD)(__FUNCTION__)(": Failed to decode message payload")
             .Flush();
+        disconnect();
```
justusranvier, line 696: This is only needed in the verify state.

src/blockchain/p2p/bitcoin/Peer.hpp
```diff
@@ -120,6 +120,9 @@ class Peer final : public p2p::implementation::Peer
     const std::set<p2p::Service> local_services_;
     std::atomic<bool> relay_;
     Request get_headers_;
+    bool services_network_;
+    bool services_limited_;
+    bool services_cfilters_;
```
justusranvier, line 125: These member variable are redundant. The parent class has a protected member variable called addresses_ which stores this information.

src/blockchain/client/HeaderOracle.cpp
FellowTraveler: Done

src/blockchain/client/HeaderOracle.hpp
FellowTraveler: Done

src/blockchain/p2p/bitcoin/Peer.cpp
FellowTraveler: Done

src/blockchain/p2p/bitcoin/Peer.hpp
FellowTraveler: Done

src/blockchain/p2p/Peer.cpp
```diff
@@ -357,6 +392,51 @@ auto Peer::handshake() noexcept -> void
     if (disconnect) { this->disconnect(); }
 }
 
+
+auto Peer::subscribe() noexcept -> void
```
justusranvier, line 396: None of the actions in the start_subscribe functions wait on network activity and happen in the same thread.

There's no need for a while loop to wait on it to complete.

PR #1516, opened 2020-10-29
Support building as a subproject

justusranvier:


randy-waterhouse:
Probably should have used autotools.

PR #1526, opened 2020-11-15
Added ProfileId to Context

FellowTraveler:


src/api/Core.cpp
```diff
@@ -277,7 +280,7 @@ auto Core::make_master_key(
 
     auto reason = factory.PasswordPrompt("Choose a master password");
     auto masterPassword = factory.Secret(0);
-    caller.AskTwice(reason, masterPassword);
+    caller.AskTwice(reason, masterPassword, "");
```
justusranvier, line 283: Is it possible to explicitly pass in the profile id here?

src/api/Core.hpp
```diff
@@ -82,7 +82,8 @@ class Core : virtual public api::internal::Core,
         const opentxs::Lock& lock,
         Secret& secret,
         const PasswordPrompt& reason,
-        const bool twice) const -> bool final;
+        const bool twice,
+        const std::string& key="") const -> bool final;
```
justusranvier, line 86: The default argument isn't necessary in this class since it's defined in the parent class.

tests/context/Test_PasswordCallback.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include "opentxs/core/crypto/OTCaller.hpp"
+#include "opentxs/core/crypto/OTCallback.hpp"
+
+std::string profile_id_{};
+std::string nym_id_{};
+std::string server_id_{};
+
+#define TEST_PASSWORD "blah foo blah foo blah"
+
+class TestCallback : public opentxs::OTCallback
```
justusranvier, line 40: This test does not verify that the key matches the profile id

tests/context/Test_PasswordCallback.cpp
```diff
@@ -0,0 +1,108 @@
...
+    const auto nym = client.Wallet().Nym(reason);
+
+    ASSERT_TRUE(nym);
+    nym_id_ = nym->ID().str();
+    EXPECT_FALSE(nym_id_.empty());
+
+    ot::Cleanup();
+}
+
+TEST(PasswordCallback, load)
```
justusranvier, line 86: In addition to this check, the test needs to verify that nyms fail to load if the incorrect password is supplied.

src/api/Core.cpp
FellowTraveler: Done.

src/api/Core.hpp
FellowTraveler: Done

tests/context/Test_PasswordCallback.cpp
FellowTraveler: The "create" test has this line:
    profile_id_ = otx.ProfileId();
(assigns to "member").

Then the "load" test compares the profileId on the context to that member:
    const auto profile_id = otx.ProfileId();
    EXPECT_EQ(profile_id, profile_id_);

What am I missing?

tests/context/Test_PasswordCallback.cpp
FellowTraveler: Done.

PR #1718, opened 2022-01-04
Added a unit test for AmountValidator.

tomlandrum:


src/core/display/Scale_imp.hpp
```diff
@@ -217,6 +225,8 @@ struct Scale::Imp {
             effMax = default_max_.value_or(0u);
         }
 
+        if (effMax > absolute_max_) effMax = absolute_max_;
```
justusranvier, line 228: why not use std::min here?

tests/core/Test_DisplayScale.cpp
```diff
@@ -75,6 +77,42 @@ TEST(DisplayScale, usd)
...
+        const auto largest_whole_number_plus_one =
+            usd.Import(u8"115792089237316195423570985008687907853"
+                       u8"269984665640564039457584007913129639936");
+    } catch (std::out_of_range&) {
+        EXPECT_TRUE(true);
+    } catch (...) {
+        EXPECT_TRUE(false);
+    }
+
+    const auto largest_fraction = opentxs::unsigned_amount(
```
justusranvier, line 102: isn't this the smallest fraction instead of the largest?

tests/core/Test_DisplayScale.cpp
```diff
@@ -75,6 +77,42 @@ TEST(DisplayScale, usd)
...
+        largest_fraction);
+
+    EXPECT_EQ(
+        usd.Format(largest_fraction, 0, 0, 64),
+        std::string{
+            u8"$0.000\u202F000\u202F000\u202F000\u202F000\u202F000\u202F05"});
+
     const auto commaTest = std::vector<std::pair<opentxs::Amount, std::string>>{
         {00, u8"$0"},
         {1, u8"$1"},
```
justusranvier, line 118: We should have a test for 115792089237316195423570985008687907853269984665640564039457584007913129639935.00000000000000000005

PR #1727, opened 2022-01-10
removed otx::String::Format as part of bug 57

str0g:


include/opentxs/core/String.hpp
```diff
@@ -118,6 +111,7 @@ class OPENTXS_EXPORT String
     minus 1. For example, if the max is set to 10, then the valid range is 0..9.
     Therefore 9 (10 minus 1) is where the nullptr terminator goes. */
     virtual void Set(const char* data, std::uint32_t enforcedMaxLength = 0) = 0;
+    virtual void Set(const std::string&) = 0;
```
justusranvier, line 114: Instead of adding this overload use the version above and can call c_str() on the input string

include/opentxs/util/StandardFileNames.hpp
```diff
@@ -0,0 +1,32 @@
...
+UnallocatedCString getFilenameCrn(TransactionNumber number);
+
+UnallocatedCString getFilenameSuccess(const char* filename);
+
+UnallocatedCString getFilenameFail(const char* filename);
+
+UnallocatedCString getFilenameError(const char* filename);
+
+}
+#endif  // OPENTXS_STANDARDFILENAMES_HPP
```
justusranvier, line 32: These are not function that we will add to the public API so this is the wrong location for this header. Since these function are for internal use then the header should be in src/internal/util/

However they don't belong in a new header in the first place. The right location for these functions is as members of the api::Legacy class.

src/otx/blind/mint/Imp.cpp
```diff
@@ -39,6 +39,36 @@
...
+
+    std::string tmp;
+    tmp.reserve(unit_id.length() + app.length());
+
+    tmp.append(unit_id);
+    tmp.append(app);
+
+    return tmp;
+}
+
```
justusranvier, line 71: These functions should be members of the opentxs::api::Legacy class.

The argument types should be const identifier::UnitDefinition& and const identifier::Notary& as appropriate.

Also follow the project conventions of using trailing return types and the UnallocatedCString alias instead of std::string

src/otx/common/Ledger.cpp
```diff
@@ -50,6 +50,19 @@
...
+std::string fast_concat(const std::string& notary_id, const std::string& path_separator, const std::string& id) {
+    std::string tmp;
+    tmp.reserve(notary_id.length() + path_separator.length() + id.length());
+    tmp.append(notary_id);
+    tmp.append(path_separator);
+    tmp.append(id);
+
+    return tmp;
+}
+
```
justusranvier, line 65: This looks like duplication of the functions added in otx::blind::mint. All the same comments there apply here too.

justusranvier (changes requested):

PR #1768, opened 2022-02-21
Bug-295 - tests for gtest/gmock env

spamaa:
These are tests, that checks does gtest/gmock basic features works. When we will use a new kind of gmock/gtest feature, we should add a test that verifies if it works at all enviroments.

### DIFF
--- a/tests/dummy/Interfaces/DummyInterface.hpp
+++ b/tests/dummy/Interfaces/DummyInterface.hpp
@@ -1,0 +1,28 @@
+// Copyright (c) 2010-2022 The Open-Transactions developers
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#pragma once
+
+namespace opentxs::dummyinterface
+{
+class Interface {
+public:
+    virtual bool someFunction(const int &input) = 0;
+    virtual bool someConstFunction(const int& input) = 0;
+    virtual bool someFunctionWithNoExcept(const int &input) noexcept= 0;
+    virtual bool someConstFunctionWithNoExcept(const int &input) const noexcept = 0;
+
+    virtual ~Interface() = default;
+
+protected:
+    Interface() = default;
+
+private:
+    Interface(const Interface &) = delete;
+    Interface(Interface&&) = delete;
+    Interface& operator=(const Interface&) = delete;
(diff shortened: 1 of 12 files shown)


tests/dummy/Interfaces/DummyInterface.hpp
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2010-2022 The Open-Transactions developers
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#if !defined (__TEST_DUMMY_INTERFACES_DUMMYINTERFACE_HPP_INCLUDED__)
```
justusranvier, line 6: Use #pragma once instead of header guards

src/CMakeLists.txt
```diff
@@ -405,6 +405,9 @@ if(OT_INSTALL_LIBRARY_DEPENDENCIES)
   endif()
 
   if(OPENTXS_BUILD_TESTS)
+    # TEST_BEGIN
+
+    # TEST_END
```
justusranvier, line 410: What are these comments placeholders for?

justusranvier (changes requested):
See [this comment](https://github.com/Open-Transactions/opentxs-cmake/pull/1#issuecomment-1047025184) for alternatives to universally disabling Weffc++ in all our projects.

The best approach is to use the #pragma GCC technique since modifying the compiler flags for a specific file makes unity builds less effective but if that is too unwieldy turning off Weffc++ on the specific test files that use gmock is also acceptable.

tests/dummy/Interfaces/DummyInterface.hpp
spamaa: #pragma once  works only at compilers that support it, and usually ignored by compilers that do not support it. If it is not supported, compilers usually will not give even warning. At the other hand, define is more error prone. You can mismatch defines, but it almost always produces easy to detect compilation errors. Hence I prefer the define guards. Code is more robust and cost almost nothing. I will change it to keep consistent with remaining code.

src/CMakeLists.txt
spamaa: Sorry, I was using some debug code and it was just for clear identification what is a debug code and what isn't. I have forgot to delete it.

spamaa: done

PR #1772, opened 2022-02-24
Bug-249 - remove not effective command line options

spamaa:
Moved settings that are unconditionally explicit set, and which as I have understood necessary to set to specific values, into initialization section.


spamaa:
I think, there is one path, that these changes will not be transparent. I will not commit these changes without tests, which we do not have.

PR #1797, opened 2022-03-17
Bug 87, Ot.hpp unit tests + adding/linking gmock

bartalke:


tests/ui/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ else()
 endif()
 
 add_opentx_test(unittests-opentxs-ui-account-tree Test_AccountTree.cpp)
+set_tests_properties(unittests-opentxs-ui-account-tree PROPERTIES DISABLED TRUE)
```
justusranvier, line 25: Why is this test being disabled?

bartalke: It's failing non stop, but finally I have gdb output: https://pastebin.com/tLJaMRub - https://matterfi.atlassian.net/browse/OT-239, after our last conversation I thought if it is related to CheckNum and getNymbox, then it can be disabled, unless you meant ignored ?

PR #1817, opened 2022-03-22
issue SD-15

polishcode:
- dummy commit to test GitHub Checks from circleci
- also - optimized circleci config


polishcode:
still not triggering on circleci

PR #1818, opened 2022-03-22
issue SD-62

polishcode:
- optimized circleci config
- github checks...


polishcode:
still not triggering despite having appropriate project setting in circleci :(

PR #1827, opened 2022-03-24
issue SD-64

polishcode:
- re-remove android & linux github actions
- looks like circleci is already working for us as expected (checks triggered, large unix executor resource)


justusranvier:
The github actions are still useful because forks of the main repo can optionally enable them if they wish.

PR #1836, opened 2022-03-28
OT-118. Type erasure for Reply/Request/ManagedNumber classes.

bartalke:


src/otx/Reply.cpp
```diff
@@ -36,7 +36,7 @@
 #include "serialization/protobuf/ServerReply.pb.h"
 #include "serialization/protobuf/Signature.pb.h"
 
-template class opentxs::Pimpl<opentxs::otx::Reply>;
+//template class opentxs::Pimpl<opentxs::otx::Reply>;
```
justusranvier, line 39: If we're completely done with the Pimpl template for this class then this line can be completely removed

justusranvier:
Since Circle CI stopped working and we don't know how long it will take to get that resolved I turned Actions back on for this repo, but it looks like it won't trigger unless this pull request is opened again so I'm going to close it now. Once you open it again the Actions should run and then it can be merged.

bartalke:
> Since Circle CI stopped working and we don't know how long it will take to get that resolved I turned Actions back on for this repo, but it looks like it won't trigger unless this pull request is opened again so I'm going to close it now. Once you open it again the Actions should run and then it can be merged.

Question: How can I reopen it ? or need to create a new PR ?

PR #1845, opened 2022-04-04
OT 254. Fix for Arithmetic exception - Integer divide by zero.

bartalke:


src/blockchain/node/blockoracle/Cache.cpp
```diff
@@ -134,7 +134,7 @@ auto Cache::GetBatch(allocator_type alloc) noexcept
     static constexpr auto GetTarget =
         [](auto available, auto peers, auto max, auto min) {
             return std::min(
-                max, std::min(available, std::max(min, available / peers)));
+                max, std::min(available, peers > 0 ? std::max(min, available / peers) : min));
```
justusranvier, line 137: Instead of adding a new type of construct in this statement, it would be better to use std::max(peers, 1) to stay consistent.

Also add a static_assert that checks the peers = 0 case.

bartalke: It seems that I cannot add static_assert check, because peer is not constant expression.

justusranvier: Look at the other static asserts just below this lambda definition. They verify that the function returns the correct values for various combinations of arguments.

You just need to add one or more that pass a 0 for the peers argument.

bartalke: Yeah, I realized that after I wrote it :D Please check if the new values are fine.

PR #1879, opened 2022-05-05
MT-13 Removed downcasts in Actor

wiesiekpap:
Purpose: Remove downcasts, restore the guarantee of private access in Actor subclasses, make the interactions between Actor and its subclasses more explicit and easier to follow.


justusranvier:
Using the CRTP idiom isn't causing any problems so there is no reason to change it.

PR #1882, opened 2022-05-06
MT-13 Removed CRTP, downcasting and backdoor access to private functi…

wiesiekpap:
MT-13 Removed CRTP, downcasting and backdoor access to private functions.

This has been done in preparation for the introduction of more explicit interfaces. Note that CRTP is meant to hide the implementation of some added special functionality like access from a class to a shared pointer that is owning it, and not to be used to obfuscate the essential functionality of a class. Likely, overriding the private access specifiers in the subclassses a bad practice that removes the guarantees enforced by the compiler.

### DIFF
--- a/src/api/crypto/blockchain/BalanceOracle.cpp
+++ b/src/api/crypto/blockchain/BalanceOracle.cpp
@@ -396,6 +396,10 @@ auto BalanceOracle::Imp::UpdateBalance(
     }());
 }
 
+auto BalanceOracle::Imp::do_startup() noexcept -> void {}
+
+auto BalanceOracle::Imp::do_shutdown() noexcept -> void {}
+
 auto BalanceOracle::Imp::work() noexcept -> bool { return false; }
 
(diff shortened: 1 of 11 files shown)


justusranvier:
The purpose of CRTP is to avoid virtual function calls at runtime by resolving them at compile time.

In normal inheritance the child class functions being called by Actor would be pure virtual and defined in Actor, but that introduces unnecessary runtime overhead.

CRTP should be preferred over inheritance in most implementation classes where a stable public interface is not worth the extra runtime cost.

wiesiekpap:
The kind of processing taking the bulk of cpu time is such that virtual call overhead is unlikely to be even measurable in a reliable way. This is just as well, because many of the functions involved had to be made virtual anyway, as the direct call technique only reached into the immediate subclass. The downside, however, is code that is unnecessarily hard to work with, thus prone to human errors and misinterpretation. CRTP is justified when the base class actually needs to know the type of the subclass at compile time, like when it needs to refer to a shared or weak pointer owning it.

PR #1883, opened 2022-05-06
correct history inconsistencies during wallet recovery from seed

justusranvier:


src/blockchain/block/bitcoin/Input.hpp
```diff
@@ -215,14 +222,19 @@ class Input final : public internal::Input
             std::for_each(std::begin(keys_), std::end(keys_), cb);
         }
         auto keys() const noexcept -> UnallocatedVector<crypto::Key>;
-        auto net_balance_change(const identifier::Nym& nym) const noexcept
-            -> opentxs::Amount;
+        auto net_balance_change(
+            const identifier::Nym& nym,
+            const std::size_t index,
+            const Log& log) const noexcept -> opentxs::Amount;
```
bartalke, line 228: It seems that keeping log as class member would be better if we don't want to create log instance each time when we want to log sth ;)

tomlandrum: For the logging, we use static instances instead of creating member variables.  In the case of these transaction classes, we want to pass in the log object to use the same level of logging as the calling object.

PR #1908, opened 2022-05-31
Set thread names to aid with debugging.

tomlandrum:


src/api/network/asio/Imp.cpp
```diff
@@ -277,7 +310,20 @@ auto Asio::Imp::Post(ThreadPool type, Asio::Callback cb) noexcept -> bool
...
-    boost::asio::post(pool.get(), std::move(cb));
+    boost::asio::post(
+        pool.get(),
+        [action = std::move(cb),
+         name = UnallocatedCString{"asio "}
+                    .append(print(type))
+                    .append(": ")
+                    .append(threadName),
+         type] {
+            SetThisThreadsName(name);
```
wiesiekpap, line 321: In most cases this will render unacceptably long names for Linuz, or any Posix platform.
"asioSomeType:fixed name with some embedded blanks" just won't fly here.
The "asio" is a waste of space, thread pool identification would be nice to have if we could afford it, and the original thread name with the blanks taken out would be fine, except that it does not provide uniqueness. To me, e short-ish name up to 13 characters maximu, with a provision to add a couple of digits to make it unique would do.

src/core/Shutdown.cpp
```diff
@@ -62,7 +62,7 @@ ShutdownReceiver::ShutdownReceiver(
     , callback_(zmq::ListenCallback::Factory([cb, this](auto&&) {
         if (bool(cb)) { cb(this->promise_); }
     }))
-    , socket_(zmq.SubscribeSocket(callback_))
+    , socket_(zmq.SubscribeSocket(callback_, "ShutdownReceiver"))
```
wiesiekpap, line 65: Much too long.

src/interface/rpc/RPC.cpp
```diff
@@ -225,10 +225,12 @@ RPC::RPC(const api::Context& native)
...
     }))
-    , push_receiver_(
-          ot_.ZMQ().PullSocket(push_callback_, zmq::socket::Direction::Bind))
+    , push_receiver_(ot_.ZMQ().PullSocket(
+          push_callback_,
+          zmq::socket::Direction::Bind,
+          "RPC push receiver"))
     , rpc_publisher_(ot_.ZMQ().PublishSocket())
-    , task_subscriber_(ot_.ZMQ().SubscribeSocket(task_callback_))
+    , task_subscriber_(ot_.ZMQ().SubscribeSocket(task_callback_, "RPC task"))
```
wiesiekpap, line 233: One of a number of cases of embedded blank.

src/network/ServerConnection.cpp
```diff
@@ -197,8 +198,8 @@ auto ServerConnection::Imp::activity_timer() -> void
 auto ServerConnection::Imp::async_socket(const Lock& lock) const
     -> OTZMQDealerSocket
 {
-    auto output =
-        zmq_.Context().DealerSocket(callback_, zmq::socket::Direction::Connect);
+    auto output = zmq_.Context().DealerSocket(
+        callback_, zmq::socket::Direction::Connect, "ServerConnection async");
```
wiesiekpap, line 202: Long, blanks.

PR #2188, opened 2023-10-10
Revert "check for Qt major version mismatch in dependent projects"

marcossatt:


justusranvier:
This is an important sanity check that should not be removed.

If this check is preventing a build then the configuration is invalid and removing the check won't fix it.

